PR #26, opened 2021-04-27
sklearn estimator_checks in tests

remiadon:
# Description

Inclusion of scikit-learn estimator checks (see [this page](https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.parametrize_with_checks.html))
This automates compatibility with sklearn via pytest

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

**Test Configuration**:
* OS version: Osx Sierra 10.12
* Python version: 3.9.2
* MAPIE version: 0.1.0

# Checklist:

- [X] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

## N.B
unit tests are failing, on purpose. Sklearn checks are not passing for the moment.

### DIFF
--- a/mapie/tests/test_estimators.py
+++ b/mapie/tests/test_estimators.py
@@ -216,6 +216,7 @@ def test_linreg_results(method: str) -> None:
     assert_almost_equal((preds_up-preds_low).mean(), expected_widths[method], 2)
     assert_almost_equal(coverage_score(y_reg, preds_low, preds_up), expected_coverages[method], 2)
 
+
 @parametrize_with_checks([MapieRegressor(LinearRegression())])
 def test_sklearn_compatible_estimator(estimator, check):
     check(estimator)
--- a/mapie/tests/test_estimators.py
+++ b/mapie/tests/test_estimators.py
@@ -58,6 +58,12 @@ expected_coverages = {
     "cv_minmax": 0.966
 }
 
+SKLEARN_EXCLUDED_CHECKS = {
+    "check_regressors_train",
+    "check_pipeline_consistency",
+    "check_fit_score_takes_y",
+}
+
 
 def test_optional_input_values() -> None:
     """Test default values of input parameters."""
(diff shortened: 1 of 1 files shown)


gmartinonQM:
Hi Rémi, thanks for the PR ! Actually, all unit test that fail are due to the shape of MAPIE predict output. Indeed, since the very purpose of MAPIE is to output both predictions AND prediction intervals, any call to MAPIE's predict method will output a numpy array `y_pred` of shape (n_samples, 3) with

- `y_pred[:, 0]` representing the prediction itself
- `y_pred[:, 1]` the lower bound of the prediction interval
- `y_pred[:, 2]` the upper bound of the prediction interval

This very shape of the output is not compatible with standard sklearn estimators, which explicitly require that `y.shape == y_pred.shape`.

Do you have any sklearn-compatible idea for restituting prediction intervals at prediction time ?

remiadon:
@gmartinonQM sure :)

I guess the proper way to do it is to "deactivate" the tests relative to predictions outputs shape, but at least it will be explicitly mentioned.

> Do you have any sklearn-compatible idea for restituting prediction intervals at prediction time ?

IMO there is no solution. I believe the reason why input/output shapes are tied up this way is mainly for purely predictive workflows (the output is meant to be ready to use "as-is"). In the case of MAPIE outputted intervals give more information for later processing (eg. plotting, uncertainty assessment), am I correct ?

From what I know this is not really a problem, your API diverges a little bit from sklearn's but the overall philosophy as long as it's mentionned I think it's OK :)

remiadon:
@gmartinonQM the last comment passes all checks

gmartinonQM:
> In the case of MAPIE outputted intervals give more information for later processing (eg. plotting, uncertainty assessment), am I correct ?

Correct, MAPIE is intended to extend purely predictive workflows by adding information and give insight to the final user. 

Thanks for your contribution that make the package more robust and more compliant with sklearn best practices !

Issue #128, opened 2022-01-27
MAPIE is not able to one-hot encode columns when estimator is an scikit learn pipeline with a preprocessor step


edgBR:
**Describe the bug**
Dear colleagues, I am creating a system to classify customers in 2 binary classes and then apply a regression model to one of the classes.

Some of my features are string that I obviously need to encode. In this case with one hot encoding.

**To Reproduce**

My code is as follows:

```python
from sklearnex import patch_sklearn
patch_sklearn()

from sklearn.pipeline import Pipeline
from sklearn.compose import ColumnTransformer, TransformedTargetRegressor, make_column_selector as selector 
from sklearn.preprocessing import StandardScaler, OneHotEncoder, LabelEncoder 
from sklearn.impute import SimpleImputer
#from sklearn.model_selection import cross_val_score, cross_validate
from sklearn.multioutput import RegressorChain, MultiOutputRegressor
from sklearn.feature_selection import VarianceThreshold
from sklearn.metrics import mean_absolute_error, make_scorer, mean_tweedie_deviance, auc
from sklearn.model_selection import RandomizedSearchCV, train_test_split, LeaveOneGroupOut, LeavePGroupsOut, cross_validate
from sklearn.metrics import roc_auc_score, plot_roc_curve, roc_curve, confusion_matrix, classification_report, ConfusionMatrixDisplay
from sklearn.ensemble import RandomForestClassifier, HistGradientBoostingRegressor, HistGradientBoostingClassifier

from sklearn import set_config
from mapie.regression import MapieRegressor

data_train, data_test, target_train, target_test = train_test_split(
    df.drop(columns=target_reg + target_class + METADATA_COLUMNS), 
    df[target_reg + target_class], 
    random_state=42)

categorical_columns_of_interest = ['col1', 'col2', 'col3', 'col4', 'col5', 'col6', 'col7']
numerical_columns = ml_data.drop(columns=target_reg + target_class + METADATA_COLUMNS).select_dtypes(include=np.number).columns
numerical_columns = [x for x in MY_FEATURES if x not in FEATURES_NOT_TO_IMPUTE]
numerical_columns = [x for x in numerical_columns if x not in categorical_columns_of_interest]

categorical_transformer = OneHotEncoder(handle_unknown="ignore")
numeric_transformer = Pipeline(
    steps=[
        ("imputer", SimpleImputer(strategy="mean")), 
        ("scaler", StandardScaler()),
[...]

PR #144, opened 2022-03-24
Fix typing

### LINKED ISSUE #143: Typing not working
**Describe the bug**
The command `make type-check` raise a very high number of errors.

**Expected behavior**
The command should run as-is after `git clone`

**Additional context**
It seems that there is a bug in `mapie/_typing.py` because the import statement `from np.typing import ArrayLike` raises an error (catched, but not sure this is the normal behaviour).

gmartinonQM:
# Description

Fix typing. Basically, there is now a distinction between `ArrayLike` and `NDArray`, the latter type being compatible with standard numpy methodes such as `.shape`. `ArrayLike` however, is meant to be generic (for example, nested python lists). The `_typing.py` bug is fixed.

Fixes #143 

## Type of change

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

All standard tests (see below).

# Checklist

### DIFF
--- a/environment.dev.yml
+++ b/environment.dev.yml
@@ -5,6 +5,8 @@ channels:
 dependencies:
     - bump2version=1.0.1
     - flake8=4.0.1
+    - ipykernel=6.9.0
+    - jupyter=1.0.0
     - mypy=0.910
     - numpydoc=1.1.0
     - pandas=1.3.5
--- a/environment.dev.yml
+++ b/environment.dev.yml
@@ -7,7 +7,7 @@ dependencies:
     - flake8=4.0.1
     - ipykernel=6.9.0
     - jupyter=1.0.0
-    - mypy=0.910
+    - mypy=0.941
     - numpydoc=1.1.0
     - pandas=1.3.5
     - pytest=6.2.5
--- a/mapie/classification.py
+++ b/mapie/classification.py
@@ -77,8 +77,8 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
           ``sklearn.model_selection.LeaveOneOut()``.
         - CV splitter: any ``sklearn.model_selection.BaseCrossValidator``
           Main variants are:
-            - ``sklearn.model_selection.LeaveOneOut`` (jackknife),
-            - ``sklearn.model_selection.KFold`` (cross-validation)
+          - ``sklearn.model_selection.LeaveOneOut`` (jackknife),
+          - ``sklearn.model_selection.KFold`` (cross-validation)
         - ``"prefit"``, assumes that ``estimator`` has been fitted already.
           All data provided in the ``fit`` method is then used
           to calibrate the predictions through the score computation.
--- a/mapie/classification.py
+++ b/mapie/classification.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Optional, Union, Tuple, Iterable, List
+from typing import Optional, Union, Tuple, Iterable, List, cast
 
 import numpy as np
 from joblib import Parallel, delayed
@@ -17,7 +17,7 @@ from sklearn.utils.validation import (
     _check_y,
 )
 
-from ._typing import ArrayLike
+from ._typing import ArrayLike, NDArray
 from ._machine_precision import EPSILON
 from .utils import (
     check_cv,
@@ -332,8 +332,8 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
     def _check_proba_normalized(
         self,
         y_pred_proba: ArrayLike,
-        axis: Optional[int] = 1
-    ) -> Optional[ArrayLike]:
+        axis: int = 1
+    ) -> ArrayLike:
         """
         Check if, for all the observations, the sum of
         the probabilities is equal to one.
@@ -347,7 +347,7 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
 
         Returns
         -------
-        Optional[ArrayLike] of shape (n_samples, n_classes)
+        ArrayLike of shape (n_samples, n_classes)
             Softmax output of a model if the scores all sum
             to one.
 
@@ -366,8 +366,8 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
 
     def _get_last_index_included(
         self,
-        y_pred_proba_cumsum: ArrayLike,
-        threshold: ArrayLike,
+        y_pred_proba_cumsum: NDArray,
+        threshold: NDArray,
         include_last_label: Optional[Union[bool, str]]
     ) -> ArrayLike:
         """
@@ -377,10 +377,10 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
 
         Parameters
         ----------
-        y_pred_proba_cumsum : ArrayLike of shape (n_samples, n_classes)
+        y_pred_proba_cumsum : NDArray of shape (n_samples, n_classes)
             Cumsumed probabilities in the original order.
 
-        threshold : ArrayLike of shape (n_alpha,) or shape (n_samples_train,)
+        threshold : NDArray of shape (n_alpha,) or shape (n_samples_train,)
             Threshold to compare with y_proba_last_cumsum, can be either:
 
             - the quantiles associated with alpha values when
@@ -395,7 +395,7 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
 
         Returns
         -------
-        Optional[ArrayLike] of shape (n_samples, n_classes)
+        ArrayLike of shape (n_samples, n_classes)
             Index of the last included sorted probability.
         """
         if (
@@ -434,12 +434,12 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):  # type: ignore
 
     def _add_random_tie_breaking(
         self,
-        prediction_sets: ArrayLike,
-        y_pred_index_last: ArrayLike,
-        y_pred_proba_cumsum: ArrayLike,
-        y_pred_proba_last: ArrayLike,
-        threshold: ArrayLike
-    ) -> ArrayLike:
+        prediction_sets: NDArray,
+        y_pred_index_last: NDArray,
+        y_pred_proba_cumsum: NDArray,
+        y_pred_proba_last: NDArray,
+        threshold: NDArray
+    ) -> NDArray:
         """
         Randomly remove last label from prediction set based on the
         comparison between a random number and the difference between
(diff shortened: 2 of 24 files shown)


environment.dev.yml
```diff
@@ -5,12 +5,15 @@ channels:
 dependencies:
     - bump2version=1.0.1
     - flake8=4.0.1
-    - mypy=0.910
+    - ipykernel=6.9.0
```
vincentblot28, line 8: Why adding ipykernel and jupyter to the packages ?

mapie/subsample.py
```diff
@@ -66,12 +68,12 @@ def split(
...
         ------
-        train : ArrayLike of shape (n_indices_training,)
+        train : NDArray of shape (n_indices_training,)
             The training set indices for that split.
-        test : ArrayLike of shape (n_indices_test,)
+        test : NDArray of shape (n_indices_test,)
             The testing set indices for that split.
         """
-        indices = np.arange(len(X))
+        indices = np.arange(_num_samples(X))
```
vincentblot28, line 76: why not keeping len ?

gmartinonQM: Because not all `ArrayLike` objects have a `__len__` method, so mypy complain about it.

.github/workflows/test.yml
```diff
@@ -9,43 +9,40 @@ jobs:
...
-            numpy-version: 1.18.5
-          - os: windows-latest
-            python-version: 3.8
-            numpy-version: 1.19.5
+            python-version: "3.9"
+            numpy-version: 1.21.4
+          - os: ubuntu-latest
+            python-version: "3.10"
+            numpy-version: 1.22.3
           - os: windows-latest
```
vtaquet, line 23: Why did you keep one version of Python for Windows jobs ?

mapie/_compatibility.py
```diff
@@ -0,0 +1,33 @@
...
+from ._typing import ArrayLike, NDArray
+
+
+def np_quantile_version_below_122(
+    a: ArrayLike,
+    q: ArrayLike,
+    method: str = "linear",
+    **kwargs: Any
+) -> NDArray:
+    """Wrapper of np.quantile function for numpy version < 1.22."""
```
vtaquet, line 15: No description of the input and output argument arguments ?

mapie/_compatibility.py
```diff
@@ -0,0 +1,33 @@
...
+    return np.quantile(a, q, interpolation=method, **kwargs)  # type: ignore
+
+
+def np_quantile_version_above_122(
+    a: ArrayLike,
+    q: ArrayLike,
+    method: str = "linear",
+    **kwargs: Any
+) -> NDArray:
+    """Wrapper of np.quantile function for numpy version >= 1.22."""
```
vtaquet, line 25: No description of the input and output argument arguments ?

mapie/utils.py
```diff
@@ -5,21 +5,26 @@
...
 from sklearn.utils import _safe_indexing
 
-from ._typing import ArrayLike
+from ._typing import ArrayLike, NDArray
 
 
 def check_null_weight(
-    sample_weight: ArrayLike, X: ArrayLike, y: ArrayLike
-) -> Tuple[ArrayLike, ArrayLike, ArrayLike]:
+    sample_weight: Optional[ArrayLike],
```
vtaquet, line 18: Shouldn't an optional argument be placed after mandatory ones ?

mapie/classification.py
```diff
@@ -684,9 +680,11 @@ def fit(
         X, y = indexable(X, y)
         y = _check_y(y)
         assert type_of_target(y) == "multiclass"
-        self.n_classes_ = len(set(y))
-        self.n_features_in_ = check_n_features_in(X, cv, estimator)
         sample_weight, X, y = check_null_weight(sample_weight, X, y)
+        y = cast(NDArray, y)
```
vtaquet, line 684: `y` is casted to a `NDArray` but `y` accepted by `_fit_and_predict_oof_model` is still an `ArrayLike`.  I have a related open question : shouldn't we convert all input `ArrayLike` arguments as `NDArray` at the beginning of `fit` and `predict` methods so all subsequent methods only accept and return `NDArray` arguments ?

mapie/classification.py
```diff
@@ -830,190 +835,192 @@ def predict(
...
+        alpha = cast(Optional[NDArray], check_alpha(alpha))
         check_is_fitted(self, self.fit_attributes)
         if self.image_input:
             check_input_is_image(X)
 
         # Estimate prediction sets
         y_pred = self.single_estimator_.predict(X)
-        n = self.n_samples_
-        if alpha_ is None:
+        n = len(self.conformity_scores_)
```
vtaquet, line 857: Does it make a difference ?

mapie/classification.py
```diff
@@ -830,190 +835,192 @@ def predict(
...
+
+        if alpha is None:
             return np.array(y_pred)
 
+        # Estimate of probabilities from estimator(s)
+        # In all cases : len(y_pred_proba.shape) == 3
+        # with  (n_test, n_classes, n_alpha or n_train_samples)
+        alpha_np = cast(NDArray, alpha)
```
vtaquet, line 865: Wasn't `alpha` converted to a `NDArray` line 850 ?

mapie/regression.py
```diff
@@ -368,23 +364,19 @@ def _fit_and_predict_oof_model(
...
-        - [1]: Estimator predictions on the validation fold,
-          of shape (n_samples_val,)
-        - [3]: Validation data indices,
-          of shape (n_samples_val,).
+        Tuple[RegressorMixin, NDArray, ArrayLike]
 
+        - [0]: RegressorMixin, fitted estimator
+        - [1]: NDArrayof shape (n_samples_val,),
+          estimator predictions on the validation fold.
+        - [3]: NDArray of shape (n_samples_val,),
```
vtaquet, line 378: `NDArray` or `ArrayLike` ?

mapie/regression.py
```diff
@@ -488,6 +481,8 @@ def fit(
         y = _check_y(y)
         self.n_features_in_ = check_n_features_in(X, cv, estimator)
         sample_weight, X, y = check_null_weight(sample_weight, X, y)
+        y = cast(NDArray, y)
```
vtaquet, line 484: Shouldn't we also cast `X` so all methods accept and return only `NDArray` arguments ?

mapie/regression.py
```diff
@@ -608,71 +597,74 @@ def predict(
...
         self._check_ensemble(ensemble)
-        alpha_ = check_alpha(alpha)
+        alpha = cast(Optional[NDArray], check_alpha(alpha))
         y_pred = self.single_estimator_.predict(X)
+        n = len(self.conformity_scores_)
 
         if alpha is None:
             return np.array(y_pred)
+
+        alpha_np = cast(NDArray, alpha)
```
vtaquet, line 607: `alpha` was not converted to `NDArray` in line 600 ?

.github/workflows/test.yml
gmartinonQM: Well, different python versions are already tested on linux, I wanted to save time by reducing windows testing to a single job and avoid redundancy

mapie/_compatibility.py
gmartinonQM: Initially I did copy-pasted the docstring of numpy and then it felt awkward to do this two times. I hope the docstring is explicit enough

mapie/_compatibility.py
gmartinonQM: Initially I did copy-pasted the docstring of numpy and then it felt awkward to do this two times. I hope the docstring is explicit enough

mapie/utils.py
gmartinonQM: Yes, but the subtlety here is that the type `Optional[ArrayLike]` is nothing but a synonym of `Union[ArrayLike, None]`, so the type is valid, and python knows that it is not optional since there is no default value.

mapie/classification.py
gmartinonQM: I removed the useless attributes `n_samples_` that was used nowhere and redundantly encoded in `conformity_scores_`

mapie/classification.py
gmartinonQM: No, `alpha` was cast to `Optional[NDArray]`, so that checking against `None` does satisfy mypy

mapie/regression.py
gmartinonQM: ArrayLike ! Done

mapie/regression.py
gmartinonQM: No, to `Optional[NDArray]`, so that checking against `None` is valid for mypy

mapie/classification.py
gmartinonQM: Alas no, because as some users noticed in this issue https://github.com/scikit-learn-contrib/MAPIE/issues/128, MAPIE should not convert input arrays as numpy arrays on its own, but should let the wrapped model manage arbitrary inputs. One example is a wrapped model being a pipeline treating pandas dataframe with a scikit-learn ColumnTransformer : the latter needs input to be pandas dataframe in order to record column names. Would MAPIE convert this to a numpy array, the compatibility with such pipelines would be lost.

mapie/regression.py
gmartinonQM: Alas no, because as some users noticed in this issue https://github.com/scikit-learn-contrib/MAPIE/issues/128, MAPIE should not convert input arrays as numpy arrays on its own, but should let the wrapped model manage arbitrary inputs. One example is a wrapped model being a pipeline treating pandas dataframe with a scikit-learn ColumnTransformer : the latter needs input to be pandas dataframe in order to record column names. Would MAPIE convert this to a numpy array, column names would be lost, andthe compatibility with such pipelines would be lost.

mapie/classification.py
gmartinonQM: You are right that `y` is converted to `np.array` as soon as `_check_y` is called, but mypy does not notice it because this scikit-learn function has no typing. The cast is then necessary, here or before `check_null_weight`. In this latter case, for mypy to understand that the return `y` is `NDArray`,  the signature of `check_null_weight` should changed.

vtaquet: OK understood, thanks for the explanation. So why is `y` casted here ? Because transforming it from a `pd.Series` to a `NDArray` won't affect the scikit-learn pipeline ?

mapie/classification.py
vtaquet: `mypy` is really picky !

mapie/classification.py
gmartinonQM: `y` needs to be casted to `NDArray` because MAPIE code uses typical `NDArray` methods like `reshape` further in the code.

Issue #169, opened 2022-05-30
Issue on test `test_toy_dataset_predictions` for classification


LacombeLouis:
**Describe the bug**
A problem with the test `test_toy_dataset_predictions` for the classifications when checking for the strategy: "cumulated_score_randomized_cv_crossval". This means that when ``include_last_label`` is equal to "randomized", we would get different values depending on what machine it would be run.

**To Reproduce**
1. Use a machine such as the provided Windows machine from the Workflows provided.
2. Run the tests and more specifically the `test_toy_dataset_predictions` with strategy "cumulated_score_randomized_cv_crossval".
3. See error

**Expected behavior**
Should observe that the values are different between the test predictions and what hard code written results.

**Desktop (please complete the following information):**
 - OS: macOS Monterey, version 12.4
 - MAPIE Version: 0.3.2

**Additional context**
Most likely due to a machine precision error.

PR #170, opened 2022-05-30
solve for machine precision issue

LacombeLouis:
# Description
Due to a machine precision problem, the tests would not all pass as we would be using a "<" sign as oppose to `np.less_equal` with the epsilon parameter. This is related to issue #169.

Fixes #(issue)
Changed the method to check if the value were greater or equal than by using the `np.less_equal` with an epsilon parameter as oppose to the "<" sign.

# How Has This Been Tested?

- [x] Check that all tests are passed

# Checklist

### DIFF
--- a/mapie/tests/test_classification.py
+++ b/mapie/tests/test_classification.py
@@ -530,9 +530,6 @@ class Float32OuputModel:
     def predict(self, X: NDArray, *args: Any) -> NDArray:
         return np.repeat(1, len(X))
 
-    def _get_param_names(self):
-        return ["prefit"]
-
     def get_params(self, *args: Any, **kwargs: Any):
         return {"prefit": False}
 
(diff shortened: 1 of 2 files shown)


mapie/tests/test_classification.py
```diff
@@ -530,9 +530,6 @@ def predict_proba(self, X: NDArray, *args: Any) -> NDArray:
     def predict(self, X: NDArray, *args: Any) -> NDArray:
         return np.repeat(1, len(X))
 
-    def _get_param_names(self):
```
vtaquet, line 533: Why removing this method ? No longer used ?

vtaquet (commented):
Thanks for you PR Louis ! Just a little comment : this is the introduction of the EPSILON parameter that solves the issue, not `np.less_equal`.

mapie/tests/test_classification.py
LacombeLouis: It does not seem to be needed and the coverage is not at 100% due to this. Not sure of the additional value of this mehod.

Issue #177, opened 2022-06-15
No module named `typing_extensions`


LacombeLouis:
**Describe the bug**
When you download MAPIE from GitHub, you will have the error that "No module named 'typing_extensions'". This arises from as we use `typing_extensions` in `MapieQuantileRegressor`.

**To Reproduce**
1. Try to run the command line `pip install git+https://github.com/scikit-learn-contrib/MAPIE`

**Screenshots**
<img width="832" alt="Screenshot 2022-06-15 at 16 19 27" src="https://user-images.githubusercontent.com/63096108/173871156-b9c0f6cd-70cb-4cb4-803f-aa685d0ee13a.png">

PR #178, opened 2022-06-15
TypedDict to Dict

LacombeLouis:
# Description
We had the issue that a `TypedDict` was used in `MapieQuantileRegressor` which therefore lead to the use of `typing_extensions` which is a package that was not automatically downloaded through [setup.py](https://github.com/scikit-learn-contrib/MAPIE/blob/master/setup.py). This is related to Issue #177.

Fixes #(issue)
Make all the `TypedDict` to `Dict` thereby allowing to delete the `typing_extensions` import.

# Checklist

### DIFF
--- a/mapie/tests/test_quantile_regression.py
+++ b/mapie/tests/test_quantile_regression.py
@@ -159,13 +159,6 @@ def test_no_predict_fit_estimator() -> None:
 
 def test_no_para_loss_estimator() -> None:
     """Test to check when it does not have a valid loss_name."""
-    Params = TypedDict(
-        "Params",
-        {
-            "loss_name": str,
-            "alpha_name": str
-        },
-    )
     with pytest.raises(
         ValueError,
         match=r".*The matching parameter `loss_name`*",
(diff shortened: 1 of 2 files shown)


mapie/tests/test_quantile_regression.py
```diff
@@ -159,24 +159,17 @@ def test_no_predict_fit_estimator() -> None:
 
 def test_no_para_loss_estimator() -> None:
     """Test to check when it does not have a valid loss_name."""
-    Params = TypedDict(
```
vtaquet, line 162: You can leave `TypedDict` dictionaries in the tests since they are used with `mapie-dev` conda environment that includes `typing-extensions`.

LacombeLouis: Yes, I would normally agree but in this case, for typing measures, we are also changing these since their role is to imitate an estimator. Therefore the dictionary entries needs to match with that of the class `MapieQuantileRegressor`.

vtaquet: Yes, you're right !

Issue #237, opened 2022-12-20
Year missing


cmougan:
"Yaniv Romano, Matteo Sesia and Emmanuel J. Candès. “Classification with Valid and Adaptive Coverage.” NeurIPS 202 (spotlight)."

Which NIPS?? i think its missing here a 2

PR #242, opened 2022-12-23
Update README.rst #237

cmougan:
Added the year


LacombeLouis:
Hey @cmougan,
Thank you for looking into this. This is indeed correct, there were a couple of issues with the documentation and typing that we have now fixed. I suggest you update (merge) the current branch with the latest updates made on the master branch. Then we will be able to merge your PR!

PR #255, opened 2023-01-12
FIX: impose last included proba to be one if the max score is one and…

### LINKED ISSUE #254: Large prediction set when all classes but one have probability 0
**Describe the bug**
Using MapieClassifier with method="cumulated_score" produces large prediction sets (that include all classes) for instances where all probabilities are 0 except for 1 class where it's 1 (e.g. [1,0,0] produces a set with all 3 classes).

**To Reproduce**

```Python
from mapie.classification import MapieClassifier
import numpy as np
from sklearn.datasets import load_iris
from sklearn.model_selection import train_test_split
from sklearn.ensemble import RandomForestClassifier

X,y = load_iris(return_X_y=True)
X_train, X_calib, y_train, y_calib = train_test_split(X, y, train_size=100, random_state=2)
model = RandomForestClassifier().fit(X_train, y_train)
mapie_score = MapieClassifier(model, cv="prefit", method="cumulated_score")
mapie_score.fit(X_calib, y_calib)
y_pred, y_set = mapie_score.predict(X_calib, alpha=0.05,include_last_label=False)
y_set = np.squeeze(y_set)
```

When I filter all the sets that have size==3, and look at the probability vector, I get this result:

```Python
all_classes = np.sum(y_set, axis=1) == 3
model.predict_proba(X_calib)[all_classes]
```
```
array([[0.03, 0.94, 0.03],
       [0.  , 1.  , 0.  ],
       [0.  , 0.  , 1.  ],
       [0.  , 1.  , 0.  ],
       [0.  , 1.  , 0.  ],
       [0.  , 0.  , 1.  ],
       [0.  , 1.  , 0.  ],
       [0.  , 1.  , 0.  ],
       [0.  , 0.  , 1.  ],
       [0.  , 1.  , 0.  ],
       [0.  , 0.  , 1.  ],
       [0.  , 1.  , 0.  ]])
```

**Expected behavior**
I would expect that these cases would result in prediction sets with just one class.

**Desktop (please complete the following information):**

- OS: MacOS
- Python (3.10.7)
- MAPIE Version 0.5.0

**Additional context**

This also occurred with the dry beans dataset, which has 7 classes and 13k instances. For the class that was simplest to classify, method="cumulated_score" would always produce sets containing all 7 classes. Both iris and dry beans have in common that they are "simple" in the sense that the model makes confident classifications with probability=1 for many instances.

Is that a bug, or intended behavior?
Seems like it has something to do with ties, but couldn't figure it out yet.

Thanks!

vincentblot28:
… add test accordingly

# Description

Fixes #254 

## Type of change

# How Has This Been Tested?

Modify test_cumulated_scores to take into account à example with a score to 1.

# Checklist


mapie/tests/test_classification.py
```diff
@@ -1220,7 +1220,7 @@ def test_get_last_included_proba_shape(k_lambda, strategy):
 
     mapie = MapieClassifier(estimator=clf, **STRATEGIES[strategy][0])
     include_last_label = STRATEGIES[strategy][1]["include_last_label"]
-
+    mapie.n_classes_ = len(np.unique(y))
```
LacombeLouis, line 1223: Why do you add this here?

mapie/tests/test_classification.py
```diff
@@ -528,7 +528,7 @@ def predict_proba(self, X: ArrayLike) -> NDArray:
             )
         else:
             return np.array(
-                [[0.2, 0.7, 0.1], [0.1, 0.2, 0.7], [0.3, 0.5, 0.2]]
+                [[0.2, 0.7, 0.1], [0., 1., 0.], [0., .7, 0.3]]
```
LacombeLouis, line 531: Add also maybe [0.3, 0.7, 0.], just to make sure it works in all cases.

mapie/classification.py
```diff
@@ -838,6 +838,23 @@ def _get_last_included_proba(
             axis=1
         )
 
+        zeros_scores = y_pred_proba_sorted_cumsum[:, -2, 0] == 1
+
+        if np.sum(zeros_scores) > 0:
```
LacombeLouis, line 843: I think it would be nice to add a comment here or make this into a method to clearly explain what this does. It's a bit difficult to grasp what is going on here.

mapie/tests/test_classification.py
vincentblot28: because as we only use the _get_last_included proba method, the n_classes_ attribute is not defined in MAPIE

mapie/tests/test_classification.py
vincentblot28: done

mapie/tests/test_classification.py
vincentblot28: just checked, unnecessary indeed

mapie/classification.py
vincentblot28: changed

LacombeLouis (approved):
Good fix for the issue found lately!

PR #289, opened 2023-03-23
288 error when labels dont go from 0 to l 1

vincentblot28:


mapie/classification.py
```diff
@@ -565,7 +565,7 @@ def _fit_and_predict_oof_model(
 
     def _get_true_label_cumsum_proba(
         self,
-        y: NDArray,
+        y: ArrayLike,
```
thibaultcordier, line 568: Is this change justified?

vincentblot28: yes, otherwise the typing check failes

vincentblot28: as y is directly the output of the model it may be more complexe than just an array

thibaultcordier (approved):
Approve the bug fix.

PR #314, opened 2023-06-19
311 mapieclassifier when number of labels in calib is smaller than for model fitting method score and cumulated score do not work

### LINKED ISSUE #311: [MapieClassifier] When number of labels in calib is smaller than for model fitting, method score and cumulated_score do not work

vincentblot28:
Fixes #311


mapie/LOC_synset_mapping.txt
thibaultcordier, line 1: Is it an import error?

mapie/tests/test_classification.py
```diff
@@ -1711,6 +1711,24 @@ def test_classes_prefit() -> None:
...
+    clf.fit(X, y)
+    indices_remove = np.where(y != 2)
+    X_mapie = X[indices_remove]
+    y_mapie = y[indices_remove]
+    mapie_clf = MapieClassifier(
+        estimator=clf, method="cumulated_score",
+        cv="prefit"
+    )
+    mapie_clf.fit(X_mapie, y_mapie)
+    assert (mapie_clf.label_encoder_.classes_ == np.unique(y)).all()
```
thibaultcordier, line 1729: I don't know if it's relevant to check whether the labels have the same name/number (between y and mapie_clf.predict(X).

In other words, is there any need to check that labels aren't being reindexed?

thibaultcordier (commented):
Good for me with some comments (one import error and any additional tests).

mapie/tests/test_classification.py
vincentblot28: Yes it is necessary, because otherwise at some point the could failed, this is one of the reason of the PR

mapie/LOC_synset_mapping.txt
vincentblot28: yes, removed

mapie/tests/test_classification.py
thibaultcordier: Okay, I'll let you add it.

thibaultcordier (approved):
Okay to merge it to main branch.

PR #324, opened 2023-07-05
ENH: prediction sets for binary classification

### LINKED ISSUE #322: [DOCUMENTATION] Add Binary Classification Section
**Is your documentation request related to a problem? Please describe.**
The MAPIE documentation doesn't mention how we deal with the case of binary classification.

Comprehensive documentation explaining the position of the MAPIE library for binary classification and the link with calibration and other complementary scikit-learn tools.

### LINKED ISSUE #323: [ENHANCEMENT] Prediction Sets for Binary Classification
MAPIE offers classification methods for multi-class settings and restrictive binary setting (only single prediction).

Open prediction sets for binary classification (possible to do prediction sets with binary classification and not only single prediction).

thibaultcordier:
# Description

Open prediction sets for binary classification (possible to do prediction sets with binary classification and not only single prediction).

Fixes #323
Fixes #322

## Type of change

- Documentation and Notebook for binary classification

# How Has This Been Tested?

- [x] Duplicate multi classification tests for binary classification
- [x] Add an example for binary classification.
- [x] Add a sub-section for binary classification (predictive sets).

# Checklist


doc/theoretical_description_binary_classification.rst
```diff
@@ -0,0 +1,92 @@
...
+
+For binary classification in the distribution-free setting, there are
+different ways of handling uncertainty quantification: calibration,
+probabilistic prediction (consisting in estimating confidence intervals or CIs) 
+and set prediction (consisting in estimating prediction sets or PSs). Gupta et al.
+have established a tripod of theorems that connect these three notions for
+score-based classifier [1]. We will use their notation to present these different concepts.
+
+In MAPIE, we focus specifically on set prediction for multi-class classification
+and on calibration for binary classification.
```
LacombeLouis, line 17: The Top-Lable calibration is actually not really binary use-cases. Then, it would be better to use standard calibration techniques (as implemented on sklearn).

doc/theoretical_description_binary_classification.rst
```diff
@@ -0,0 +1,92 @@
+.. title:: Theoretical Description : contents
```
LacombeLouis, line 1: I think we should have NBR read this too.

mapie/tests/test_classification.py
```diff
@@ -1274,6 +1387,31 @@ def test_toy_dataset_predictions(strategy: str) -> None:
     )
 
 
+@pytest.mark.parametrize("strategy", [*STRATEGIES_BINARY])
```
LacombeLouis, line 1390: Do we check that prefit and split are the same? Additionally, do we add check of binary on other functions by just adding the parameters?

LacombeLouis (approved):
Good PR. Finally able to make binary predictions in MAPIE. I think these are great additions.

mapie/classification.py
```diff
@@ -232,6 +232,31 @@ def _check_parameters(self) -> None:
...
+        ----------
+        y: NDArray of shape (n_samples,)
+            Training labels.
+
+        Raises
+        ------
+        ValueError
+            If type of target is binary and method is not "score".
+        """
+        self._target_type = type_of_target(y)
```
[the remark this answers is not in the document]
thibaultcordier: `type_of_target` is a general function, isn't it? What do you mean by "general function"?

mapie/tests/test_classification.py
thibaultcordier: - Do we check that prefit and split are the same?
I don't know, but that's not what this PR is about (in the sense that this test is simply a copy and paste of the multi-class test above). It might be interesting to add it to the questions if that's the case :)
- Do we add check of binary on other functions by just adding the parameters?
If this test works for binary, then by extension, all other tests on multi-class will work for binary.

doc/theoretical_description_binary_classification.rst
thibaultcordier: Yes, in this case, I will adapt this paragraph.

doc/theoretical_description_binary_classification.rst
```diff
@@ -0,0 +1,92 @@
...
+probabilistic prediction (consisting in estimating confidence intervals or CIs) 
+and set prediction (consisting in estimating prediction sets or PSs). Gupta et al.
+have established a tripod of theorems that connect these three notions for
+score-based classifier [1]. We will use their notation to present these different concepts.
+
+In MAPIE, we focus specifically on set prediction for multi-class classification
+and on calibration for binary classification.
+
+Although set prediction is possible for binary classification, we don't recommend using this setting.
+Here is an argument from Gupta et al [1]:
```
thibaultcordier, line 20:
```suggestion
In MAPIE, we focus specifically on set prediction for multi-class classification
and on calibration for binary classification (via scikit-learn).

Although set prediction is possible for binary classification, we don't recommend using this setting. Instead, we recommend the use of calibration (see more details in the
Calibration section or by using the
:class:`~sklearn.calibration.CalibratedClassifierCV` proposed by scikit-learn
or :class:`~mapie.calibration.MapieCalibrator` proposed in MAPIE
even if top-Label calibration is actually not really binary use-cases).
Here is an argument from Gupta et al [1]:
```

doc/theoretical_description_binary_classification.rst
thibaultcordier: I make a suggestion in another comment.

doc/theoretical_description_binary_classification.rst
thibaultcordier: @LacombeLouis is it better in your opinion?

mapie/classification.py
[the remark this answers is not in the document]
LacombeLouis: Like why do we check if it's binary and then what type of target it is? Can't we do it all in one function?

mapie/tests/test_classification.py
LacombeLouis: - We do check for other multi-class problems that the split and prefit are the same.
- I don't think we need to check that multi-class works in the use-case of binary. But that the tests that are done for multi-class, all that can be, should be replicated for the binary use-case.

mapie/classification.py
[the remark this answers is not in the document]
thibaultcordier: If I understand well, you want to merge the following part in lines 1064-1065, doesn't it?
```python
check_classification_targets(y)
self._check_binary_target(y)
```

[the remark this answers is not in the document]
LacombeLouis: I think we should just do a general function that does:
`check_type_target()`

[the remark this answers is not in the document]
LacombeLouis: I think we should just do a general function that does:
`check_type_target()`

PR #337, opened 2023-08-04
Add tests for check_cv method with and without random_state

### LINKED ISSUE #336: Cross-validation not seeded implies wrong Prediction Intervals
Hello, I encountered a problem in MAPIE. When we use MapieRegressor with J+ or CV+ without any random state, then the Kfold and LeaveOneOut methods of sklearn are not seeded. It means that when computing the $R_i^{LOO}$, it happens that we compute them with which used the $i$th observation as training point. Here is a reproducible example to show that, without seed, when calling `cv.split(X)` at two different moments leads to different split of the data (which is the case in EnsembleRegressor)

```python
from mapie.regression import MapieRegressor
from sklearn.datasets import make_regression
from sklearn.model_selection import train_test_split
X, y = make_regression(10)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=.2)

mapie = MapieRegressor(method="plus", cv=5)
mapie.fit(X_train, y_train)
for train_index, _ in mapie.estimator_.cv.split(X_train):
    print("Train index: ", train_index)
    
for _, cal_index in mapie.estimator_.cv.split(X_train):
    print("Cal index: ", cal_index)
```

thibaultcordier:
# Description

The modification proposes to resolve the bug by imposing a default value for random_state in check_cv method tests, so that the method retains split behavior even if random_state is set to None.

Fixes #336

## Type of change

# How Has This Been Tested?

- [ ] Test if check_cv retains split behavior if random_state is set to None.
- [ ] Test if check_cv retains split behavior if random_state is set a certain value.

# Checklist

- [ ] I have read the [contributing guidelines](https://github.com/simai-ml/MAPIE/blob/master/CONTRIBUTING.rst)
- [ ] Linting passes successfully : `make lint`
- [ ] Typing passes successfully : `make type-check`
- [ ] Unit tests pass successfully : `make tests`
- [ ] Coverage is 100% : `make coverage`
- [ ] Documentation builds successfully : `make doc`

### DIFF
--- a/mapie/tests/test_utils.py
+++ b/mapie/tests/test_utils.py
@@ -7,17 +7,17 @@ import pytest
 from numpy.random import RandomState
 from sklearn.datasets import make_regression
 from sklearn.linear_model import LinearRegression
+from sklearn.model_selection import BaseCrossValidator
 from sklearn.utils.validation import check_is_fitted
 
 from mapie._typing import ArrayLike, NDArray
 from mapie.regression import MapieQuantileRegressor
 from mapie.utils import (check_alpha, check_alpha_and_n_samples,
-                         check_binary_zero_one, check_lower_upper_bounds,
-                         check_n_features_in, check_n_jobs, check_null_weight,
-                         check_number_bins, check_split_strategy,
-                         check_verbose, compute_quantiles, fit_estimator,
-                         get_binning_groups)
-
+                         check_binary_zero_one, check_cv,
+                         check_lower_upper_bounds, check_n_features_in,
+                         check_n_jobs, check_null_weight, check_number_bins,
+                         check_split_strategy, check_verbose,
+                         compute_quantiles, fit_estimator, get_binning_groups)
 
 X_toy = np.array([0, 1, 2, 3, 4, 5]).reshape(-1, 1)
 y_toy = np.array([5, 7, 9, 11, 13, 15])
@@ -32,8 +32,8 @@ ALPHAS = [
     np.array([0.05, 0.1, 0.2]),
 ]
 
-
-prng = RandomState(1234567890)
+random_state = 1234567890
+prng = RandomState(random_state)
 y_score = prng.random(51)
 y_scores = prng.random((51, 5))
 y_true = prng.randint(0, 2, 51)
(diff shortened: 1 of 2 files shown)


mapie/tests/test_utils.py
```diff
@@ -410,3 +410,33 @@ def test_change_values_zero_one() -> None:
...
+    for train_index, _ in cv.split(X):
+        train_indices_1.append(train_index)
+    for train_index, _ in cv.split(X):
+        train_indices_2.append(train_index)
+
+    for i in range(cv.get_n_splits()):
+        np.testing.assert_allclose(train_indices_1[i], train_indices_2[i])
+
+
+@pytest.mark.parametrize("cv", [5, "split"])
```
LacombeLouis, line 430: I think we should also test for `cv=-1`.

mapie/utils.py
```diff
@@ -171,6 +171,9 @@ def check_cv(
     ValueError
         If the cross-validator is not valid.
     """
+    if random_state is None:
```
LacombeLouis, line 174: You need to add `None` to one of the possible options. In the `raise ValueError()`.

.readthedocs.yml
```diff
@@ -1,10 +1,11 @@
 version: 2
 
 build:
-  image: latest
+  os: ubuntu-22.04
```
LacombeLouis, line 4: Why this change?

LacombeLouis (approved):
I think this fixes the problem. I just think some small details need to be added.

.readthedocs.yml
thibaultcordier: Initially, this was a test because readthedocs failed to compile the documentation. I solved the problem in another PR, but it was the same changes.

mapie/tests/test_utils.py
```diff
@@ -410,3 +410,33 @@ def test_change_values_zero_one() -> None:
...
+    cv = check_cv(cv, random_state=None)
+
+    train_indices_1, train_indices_2 = [], []
+    for train_index, _ in cv.split(X):
+        train_indices_1.append(train_index)
+    for train_index, _ in cv.split(X):
+        train_indices_2.append(train_index)
+
+    for i in range(cv.get_n_splits()):
+        np.testing.assert_allclose(train_indices_1[i], train_indices_2[i])
```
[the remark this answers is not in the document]
thibaultcordier: Because I see this way to test the equality in other files, but yes, I think it does the same thing.

mapie/tests/test_utils.py
thibaultcordier: I will test it, thanks :)

mapie/tests/test_utils.py
```diff
@@ -410,3 +410,33 @@ def test_change_values_zero_one() -> None:
     """Test that binary output are changed to zero one outputs."""
     array_ = check_binary_zero_one(np.array([0, 4, 4]))
     assert (np.unique(array_) == np.array([0, 1])).all()
+
+
+@pytest.mark.parametrize("cv", [5, "split"])
```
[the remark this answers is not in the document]
thibaultcordier: I will test it, thanks :)

mapie/utils.py
LacombeLouis: Ah no no. All good 😀

mapie/tests/test_utils.py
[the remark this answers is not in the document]
LacombeLouis: I think that `asset_allclose` is for decimals, which is not appropriate in this setting. In any case, I don't think it will change much.

mapie/tests/test_utils.py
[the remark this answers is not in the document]
thibaultcordier: In fact, it is not possible because with `cv=-1`, we need to have `X` (and in this test, we don't have `X`).

mapie/tests/test_utils.py
thibaultcordier: In fact, it is not possible because with `cv=-1`, we need to have `X` (and in this test, we don't have `X`).

PR #361, opened 2023-09-14
ADD: Conformal Predictive Distribution example

thibaultcordier:
# Description

Propose to add an example of how do Conformal Predictive Distribution with MAPIE.

## Type of change

- New example in the gallery.

# Checklist


examples/regression/2-advanced-analysis/plot_conformal_predictive_system.py
```diff
@@ -0,0 +1,138 @@
...
+# 3. Visualizing the cumulative distribution function
+# ---------------------------------------------------
+#
+# We now propose to visualize the cumulative distribution functions of
+# the predictive distribution in a graph in order to compare the two methods.
+
+
+nb_bins = 100
+
+plt.hist(
```
LacombeLouis, line 123: As you are plotting CDFs, your curve shouldn't go back to zero at the end.

thibaultcordier: This is a problem of matplotlib in fact...

examples/regression/2-advanced-analysis/plot_conformal_predictive_system.py
```diff
@@ -0,0 +1,138 @@
...
+    y_cdf_1[0], cumulative=True, density=True, bins=nb_bins,
+    label='Absolute Residual Score', histtype='step', alpha=0.8
+)
+plt.hist(
+    y_cdf_2[0], cumulative=True, density=True, bins=nb_bins,
+    label='Normalized Residual Score', histtype='step', alpha=0.8
+)
+plt.vlines(
```
LacombeLouis, line 131: We don't see the prediction (because they are the same), but then it's more confusing that anything else...

thibaultcordier: I propose to merge the model predictions to avoid confusion.

examples/regression/2-advanced-analysis/plot_conformal_predictive_system.py
```diff
@@ -0,0 +1,138 @@
...
+=================================================
+Make Conformal Predictive Distribution with MAPIE
+=================================================
+"""
+
+##############################################################################
+# In this advanced analysis, we propose to use MAPIE for Conformal Predictive
+# Distribution in few steps.
+#
+# [1] Schweder, T., & Hjort, N. L. (2016). Confidence, likelihood, probability
```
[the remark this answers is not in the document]
thibaultcordier: This is a way to give some references about conformal predictive distribution (it is a short example, so I don't propose a strong analysis).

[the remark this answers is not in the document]
LacombeLouis: Ah ok ok. I would just say something like this just before: "Here are some reference papers for more information about CPD."

examples/regression/2-advanced-analysis/plot_conformal_predictive_system.py
LacombeLouis: yes but solutions exist : https://stackoverflow.com/questions/10690094/cumulative-histogram-has-last-point-at-y-0

thibaultcordier: Ok, let's go then!

vincentblot28 (approved):

PR #386, opened 2023-12-13
Allow `MapieRegressor` to use any split strategy

### LINKED ISSUE #385: Allow `MapieRegressor` to use any split strategy
Split-CP is only available with `"split"` keyword or `ShuffleSplit` class of scikit-learn.

Propose more compatible split `BaseCrossValidator`/`BaseShuffleSplit`.

**Describe alternatives you've considered**
In practice, we need to use the predictions aggregation for cross-CPs and not for split-CPs. All you need to do is find out the number of splits using the `get_n_splits` method to know whether it's a split or a cross.

thibaultcordier:
# Description

Split-CP is only available with `"split"` keyword or `ShuffleSplit` class of scikit-learn.

This PR proposes to use more compatible split `BaseCrossValidator`/`BaseShuffleSplit`.

In practice, we need to use the predictions aggregation for cross-CPs and not for split-CPs. All you need to do is find out the number of splits using the `get_n_splits` method to know whether it's a split or a cross.

Fixes #385

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [x] Add test for the function `check_no_agg_cv`
- [x] Add new split method with `MapieRegressor`
- [x] Check that the other tests still work

# Checklist

### DIFF
--- a/mapie/estimator/estimator.py
+++ b/mapie/estimator/estimator.py
@@ -152,7 +152,7 @@ class EnsembleRegressor(EnsembleEstimator):
         "single_estimator_",
         "estimators_",
         "k_",
-        "use_split_method",
+        "use_split_method_",
     ]
 
     def __init__(
@@ -279,7 +279,7 @@ class EnsembleRegressor(EnsembleEstimator):
         ArrayLike of shape (n_samples_test,)
             Array of aggregated predictions for each testing sample.
         """
-        if self.method in self.no_agg_methods_ or self.use_split_method:
+        if self.method in self.no_agg_methods_ or self.use_split_method_:
             raise ValueError(
                 "There should not be aggregation of predictions "
                 f"if cv is in '{self.no_agg_cv_}', if cv >=2 "
@@ -407,7 +407,7 @@ class EnsembleRegressor(EnsembleEstimator):
         estimators_: List[RegressorMixin] = []
         full_indexes = np.arange(_num_samples(X))
         cv = self.cv
-        self.use_split_method = check_no_agg_cv(X, self.cv, self.no_agg_cv_)
+        self.use_split_method_ = check_no_agg_cv(X, self.cv, self.no_agg_cv_)
         estimator = self.estimator
         n_samples = _num_samples(y)
 
@@ -437,7 +437,7 @@ class EnsembleRegressor(EnsembleEstimator):
                     for train_index, _ in cv.split(X)
                 )
                 # In split-CP, we keep only the model fitted on train dataset
-                if self.use_split_method:
+                if self.use_split_method_:
                     single_estimator_ = estimators_[0]
 
         self.single_estimator_ = single_estimator_
@@ -490,7 +490,7 @@ class EnsembleRegressor(EnsembleEstimator):
         if not return_multi_pred and not ensemble:
             return y_pred
 
-        if self.method in self.no_agg_methods_ or self.use_split_method:
+        if self.method in self.no_agg_methods_ or self.use_split_method_:
             y_pred_multi_low = y_pred[:, np.newaxis]
             y_pred_multi_up = y_pred[:, np.newaxis]
         else:
--- a/mapie/estimator/estimator.py
+++ b/mapie/estimator/estimator.py
@@ -283,7 +283,7 @@ class EnsembleRegressor(EnsembleEstimator):
         ):
             raise ValueError(
                 "There should not be aggregation of predictions "
-                f"if cv is in '{self.no_agg_cv_}' "
+                f"if cv is in '{self.no_agg_cv_}', if cv >=2 "
                 f"or if method is in '{self.no_agg_methods_}'."
             )
         elif self.agg_function == "median":
--- a/mapie/estimator/estimator.py
+++ b/mapie/estimator/estimator.py
@@ -5,13 +5,13 @@ from typing import List, Optional, Tuple, Union, cast
 import numpy as np
 from joblib import Parallel, delayed
 from sklearn.base import RegressorMixin, clone
-from sklearn.model_selection import BaseCrossValidator, ShuffleSplit
+from sklearn.model_selection import BaseCrossValidator
 from sklearn.utils import _safe_indexing
 from sklearn.utils.validation import (_num_samples, check_is_fitted)
 
 from mapie._typing import ArrayLike, NDArray
 from mapie.aggregation_functions import aggregate_all, phi2D
-from mapie.utils import (check_nan_in_aposteriori_prediction,
+from mapie.utils import (check_nan_in_aposteriori_prediction, check_no_agg_cv,
                          fit_estimator)
 from mapie.estimator.interface import EnsembleEstimator
 
@@ -278,7 +278,9 @@ class EnsembleRegressor(EnsembleEstimator):
         ArrayLike of shape (n_samples_test,)
             Array of aggregated predictions for each testing sample.
         """
-        if self.method in self.no_agg_methods_ or self.cv in self.no_agg_cv_:
+        if self.method in self.no_agg_methods_ or (
+            check_no_agg_cv(self.cv, self.no_agg_cv_)
+        ):
             raise ValueError(
                 "There should not be aggregation of predictions "
                 f"if cv is in '{self.no_agg_cv_}' "
@@ -434,8 +436,9 @@ class EnsembleRegressor(EnsembleEstimator):
                     )
                     for train_index, _ in cv.split(X)
                 )
-            if isinstance(cv, ShuffleSplit):
-                single_estimator_ = estimators_[0]
+                # In split-CP, we keep only the model fitted on train dataset
+                if check_no_agg_cv(cv, self.no_agg_cv_):
+                    single_estimator_ = estimators_[0]
 
         self.single_estimator_ = single_estimator_
         self.estimators_ = estimators_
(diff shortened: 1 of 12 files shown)


mapie/utils.py
```diff
@@ -208,6 +209,37 @@ def check_cv(
...
+        True if `cv` is a split equivalent / non-aggregated cv method.
+    """
+    if isinstance(cv, str):
+        return cv in no_agg_cv_array
+    elif isinstance(cv, int):
+        return cv == 1
+    try:
+        return cv.get_n_splits() == 1
+    except Exception:
+        return False
```
vincentblot28, line 240: We should avoid using try/except conditions
```suggestion
    if hasattr(cv, "get_n_splits"):
        return cv.get_n_splits()
    else:
        return False    
```

mapie/estimator/estimator.py
```diff
@@ -278,7 +278,9 @@ def _aggregate_with_mask(
         ArrayLike of shape (n_samples_test,)
             Array of aggregated predictions for each testing sample.
         """
-        if self.method in self.no_agg_methods_ or self.cv in self.no_agg_cv_:
+        if self.method in self.no_agg_methods_ or (
+            check_no_agg_cv(self.cv, self.no_agg_cv_)
+        ):
```
vincentblot28, line 283: In the error message, add "or if cv >=2"

mapie/estimator/estimator.py
```diff
@@ -434,8 +436,9 @@ def fit(
                     )
                     for train_index, _ in cv.split(X)
                 )
-            if isinstance(cv, ShuffleSplit):
-                single_estimator_ = estimators_[0]
+                # In split-CP, we keep only the model fitted on train dataset
+                if check_no_agg_cv(cv, self.no_agg_cv_):
```
vincentblot28, line 440: check_no_agg_cv is used multiple times and always return the same results. I suggest that you define its value once and for all : 
```python
self.cv_has_no_agg = check_no_agg_cv(cv, self.no_agg_cv)
```

mapie/tests/test_utils.py
```diff
@@ -474,3 +475,17 @@ def test_check_cv_same_split_no_random_state(cv: BaseCrossValidator) -> None:
...
+    np.testing.assert_almost_equal(check_no_agg_cv(1, array), True)
+    np.testing.assert_almost_equal(check_no_agg_cv(2, array), False)
+    cv = "split"
+    np.testing.assert_almost_equal(check_no_agg_cv(cv, array), True)
+    cv = KFold(5)
+    np.testing.assert_almost_equal(check_no_agg_cv(cv, array), False)
+    cv = ShuffleSplit(1)
+    np.testing.assert_almost_equal(check_no_agg_cv(cv, array), True)
+    cv = ShuffleSplit(2)
+    np.testing.assert_almost_equal(check_no_agg_cv(cv, array), False)
```
vincentblot28, line 491:
```suggestion
@pytest.mark.parametrize(
    "cv_result", [
        (1, True), (2, False),
        ("split", True), (Kfold(5), False),
        (ShuffleSplit(1), True),
        (ShuffleSplit(2), False)
    ]
)
def test_check_no_agg_cv(cv_result: Tuple) -> None:
    array = ["prefit", "split"]
    cv, result = cv_result
    np.testing.assert_almost_equal(check_no_agg_cv(cv, array), result)
```

mapie/estimator/estimator.py
thibaultcordier: This is also the case when we test `self.method in self.no_agg_methods_` for example. For the sake of clarity, I prefer to leave it without using the object attribute and avoid multiplying object attributes that are unnecessary for the user in particular.

mapie/utils.py
thibaultcordier: We should keep `return cv.get_n_splits() == 1` in this case.

vincentblot28: Indeed, I forgot it

thibaultcordier: The other problem is that `get_n_splits` sometimes needs additional parameters. We have to throw an exception in all cases.

> E           TypeError: LeaveOneOut.get_n_splits() missing 1 required positional argument: 'X'

thibaultcordier: @vincentblot28, I think I should keep the first proposal

vincentblot28: If we know the different possible cases, we must check them independantly, otherwise we do not clearly control what happens in the try/except condition (there are no try/except in MAPIE and we should keep it this way I think)

mapie/tests/test_utils.py
thibaultcordier: Thank you for this shorter, more readable proposal. I've just made the changes.

mapie/estimator/estimator.py
vincentblot28: fine

mapie/utils.py
thibaultcordier: We don't know all the possibilities, because users can create their own personalised CVs. Also, this would involve hard coding, which is harder to maintain. I suggest using try/except for good reason.

vincentblot28: In all cases, we must be shure of the number of split, I don't think it's a good behaviour to consider that the number of split is not equal to 1 as soon as `cv` has no attribut `get_n_split``

```suggestion
    if isinstance(cv, LeaveOneOut):
        return False
    if hasattr(cv, "get_n_split"):
        return cv.get_n_split() == 1 
    else:
        raise Exception("Your `cv` argument is not an `int` nore a `str` it should
        have a `get_n_split` method")
```

thibaultcordier: After discussion, we're going to opt to externalise the "use_split_method" value with X dependency.

mapie/regression/regression.py
```diff
@@ -315,7 +315,7 @@ def _check_agg_function(
                 "You need to specify an aggregation function when "
                 f"cv's type is in {self.cv_need_agg_function_}."
             )
-        elif (agg_function is not None) or (self.cv in self.no_agg_cv_):
+        elif agg_function is not None:
```
thibaultcordier, line 318: I'm removing this restriction to make the code simpler (unnecessary if condition) because the `_aggregate_with_mask` in `estimator.py` does the job and the `_check_agg_function` returns a default value in all cases.

mapie/estimator/estimator.py
```diff
@@ -278,10 +279,10 @@ def _aggregate_with_mask(
         ArrayLike of shape (n_samples_test,)
             Array of aggregated predictions for each testing sample.
         """
-        if self.method in self.no_agg_methods_ or self.cv in self.no_agg_cv_:
+        if self.method in self.no_agg_methods_ or self.use_split_method:
             raise ValueError(
                 "There should not be aggregation of predictions "
-                f"if cv is in '{self.no_agg_cv_}' "
+                f"if cv is in '{self.no_agg_cv_}', if cv >=2 "
```
LacombeLouis, line 285: Why do we return `self.no_agg_cv_` in the error message when we use `self.use_split_method` in the if condition?

mapie/regression/time_series_regression.py
```diff
@@ -367,8 +367,9 @@ def predict(
         self.lower_quantiles_ = lower_quantiles
         self.higher_quantiles_ = higher_quantiles
 
-        if self.method in self.no_agg_methods_ \
-                or self.cv in self.no_agg_cv_:
+        if self.method in self.no_agg_methods_ or (
```
LacombeLouis, line 370: The brackets here are strange. Was this done on purpose?

mapie/estimator/estimator.py
```diff
@@ -152,6 +152,7 @@ class EnsembleRegressor(EnsembleEstimator):
         "single_estimator_",
         "estimators_",
         "k_",
+        "use_split_method",
```
LacombeLouis, line 155: Why are we adding this in the `fit_attributes`. I feel like this might create problems for people that have a custom algorithm.

LacombeLouis (commented):
Good PR! Thank you @thibaultcordier 😀
I am a little worried about adding another `fit_attributes` for custom algorithms. What do you think?

mapie/estimator/estimator.py
thibaultcordier: Because `self.use_split_method` is computed based on `self.no_agg_cv_` in other part of the code.

mapie/regression/time_series_regression.py
thibaultcordier: I could change this.

mapie/estimator/estimator.py
thibaultcordier: There won't be any problems with custom algo :)

mapie/utils.py
```diff
@@ -208,6 +209,41 @@ def check_cv(
...
+    Returns
+    -------
+    bool
+        True if `cv` is a split equivalent / non-aggregated cv method.
+    """
+    if isinstance(cv, str):
+        return cv in no_agg_cv_array
+    elif isinstance(cv, int):
+        return cv == 1
+    if hasattr(cv, "get_n_splits"):
```
vincentblot28, line 241: elif

mapie/utils.py
```diff
@@ -208,6 +209,41 @@ def check_cv(
...
+        return cv.get_n_splits(X) == 1
+    else:
```
vincentblot28, line 243: shouldn't we raise an error if none of the above conditions is satsitified ?

mapie/tests/test_utils.py
```diff
@@ -474,3 +475,18 @@ def test_check_cv_same_split_no_random_state(cv: BaseCrossValidator) -> None:
...
+
+
+@pytest.mark.parametrize(
+    "cv_result", [
+        (1, True), (2, False),
+        ("split", True), (KFold(5), False),
+        (ShuffleSplit(1), True),
+        (ShuffleSplit(2), False),
+        (object(), False)
+    ]
```
vincentblot28, line 487:
```suggestion
    "cv_result", [
        (1, True), (2, False),
        ("split", True), (KFold(5), False),
        (ShuffleSplit(1), True),
        (ShuffleSplit(2), False),
        (object(), False)
        (LeaveOneOut, False)
    ]
```

vincentblot28 (approved):
Everything is fine, let's merge !

PR #427, opened 2024-04-09
ENH: move get_true_label_position to utils in classification

BaptisteCalot:
# Description

1- Move get_true_label_position function from classification.py to a new utils file
2- Add a test for this function

# How Has This Been Tested?

1- Check if the return of my function is correct
2- Check if the shape of my function's return is correct


mapie/tests/test_classification.py
```diff
@@ -1865,8 +1869,7 @@ def test_get_true_label_position(
     y_pred_proba = y_true_proba_place[1]
     place = y_true_proba_place[2]
 
-    mapie = MapieClassifier(random_state=random_state)
-    found_place = mapie._get_true_label_position(y_pred_proba, y_true)
+    found_place = get_true_label_position(y_pred_proba, y_true)
```
vincentblot28, line 1872: This test should be implemented directly in test_utils_classification_conformity_scores

mapie/tests/test_utils_classification_conformity_scores.py
```diff
@@ -0,0 +1,21 @@
...
+    y = np.reshape(
+        y, (-1, 1)
+    )  # add in order to have shape of form (4,1) instead of (4,)
+
+    position = get_true_label_position(y_pred_proba, y)
+
+    expected_position = np.array([[0], [0], [1], [1]])
+
+    assert np.array_equal(position, expected_position)
+    assert position.shape == y.shape
```
vincentblot28, line 21: Your tests must test only one thing at a time. For the test on the shape, you can take some random np arrays with larger sizes

LacombeLouis:
Hey @BaptisteCalot,
Thank you for this PR! For the moment it doesn't pass some tests, but we have fixed the issue. So you can git merge upstream/main and everything should work!
Thank you!

PR #453, opened 2024-05-27
FIX bug #452: Quantile Correction

### LINKED ISSUE #452: Coverage validity not verified in MapieRegressor when the number of calibration data is very small
**Describe the bug**
In `MapieReggresor`, the quantile calculation does not allow us to check the coverage validity when the number of calibration data is very small.

**To Reproduce**
A rigorous approach to testing the validity of the coverage was carried out by Mahdi Torabi Rad (see his notebook: https://github.com/mtorabirad/MLBoost/blob/main/Episode15/Episode15Main.ipynb) in the special case where the number of calibration data is very small (6 in the following illustration).

**Expected behavior**
Use the corrected quantile ($|(n+1)(1−α)|/n$) to check the validity of the coverage for any number of calibration data.

**Screenshots**
image

thibaultcordier:
# Description

Use the corrected quantile ($|(n+1)(1−α)|/n$) to check the validity of the coverage for any number of calibration data.

Fixes #452

## Type of change

# How Has This Been Tested?

- [x] Perform a FWER controlling procedure (Bonferroni) to ensure the coverage control 1−α < coverage < 1−α + 1/(n+1).
- [in next PR] Reproduce the experiment of Mahdi Torabi Rad (see his notebook: https://github.com/mtorabirad/MLBoost/blob/main/Episode15/Episode15Main.ipynb) in the special case where the number of calibration data is very small (here 6).

# Checklist


mapie/conformity_scores/conformity_scores.py
```diff
@@ -235,25 +235,29 @@ def get_quantile(
...
+        n_calib = np.min(np.sum(~np.isnan(conformity_scores), axis=axis))
+        signed = 1-2*reversed
+        alpha_ref = (1-2*alpha_np)*reversed + alpha_np
+
+        quantile = signed * np.column_stack([
             np_nanquantile(
-                conformity_scores.astype(float),
-                _alpha,
+                signed * conformity_scores.astype(float),
+                np.ceil(_alpha*(n_calib+1))/n_calib,
```
LacombeLouis, line 255: Why don't we initialize this before, creating the list of all the new alphas.

mapie/regression/regression.py
```diff
@@ -628,7 +628,7 @@ def predict(
 
             alpha_np = cast(NDArray, alpha)
             if not allow_infinite_bounds:
-                n = len(self.conformity_scores_)
+                n = np.sum(~np.isnan(self.conformity_scores_))
```
LacombeLouis, line 631: Why do we only calculate from the scores that aren't `np.nan`?

mapie/tests/test_conformity_scores.py
```diff
@@ -382,18 +382,17 @@ def test_residual_normalised_prefit_get_estimation_distribution() -> None:
 @pytest.mark.parametrize("score", [AbsoluteConformityScore(),
                                    GammaConformityScore(),
                                    ResidualNormalisedScore()])
-@pytest.mark.parametrize("alpha", [[0.3], [0.5, 0.4]])
+@pytest.mark.parametrize("alpha", [[0.5], [0.5, 0.6]])
 def test_intervals_shape_with_every_score(
     score: ConformityScore,
     alpha: Any
```
LacombeLouis, line 388: Why is this `Any`?

mapie/tests/test_conformity_scores.py
```diff
@@ -382,18 +382,17 @@ def test_residual_normalised_prefit_get_estimation_distribution() -> None:
...
 ) -> None:
+    estim = LinearRegression().fit(X_toy, y_toy)
```
LacombeLouis, line 390: Why did we switch `cv="prefit"`?

mapie/tests/test_regression.py
```diff
@@ -252,6 +259,45 @@ def test_predict_output_shape(
...
+            train_test_split(X_cal_test, y_cal_test, test_size=n_test)
+        mapie_reg.fit(X_cal, y_cal)
+        _, y_pis = mapie_reg.predict(X_test, alpha=1-delta)
+        coverage = \
+            regression_coverage_score(y_test,  y_pis[:, 0, 0], y_pis[:, 1, 0])
+        cov_list.append(coverage)
+
+    # Here we are testing whether the average coverage is statistically
+    # less than the target coverage.
+    from scipy.stats import ttest_1samp
```
LacombeLouis, line 292: Why don't we import this at the top?

mapie/tests/test_regression.py
```diff
@@ -252,6 +259,45 @@ def test_predict_output_shape(
...
+    mean_low, mean_up = delta, delta + 1/(n_calib+1)
+    _, pval_low = ttest_1samp(cov_list, popmean=mean_low, alternative='less')
+    _, pval_up = ttest_1samp(cov_list, popmean=mean_up, alternative='greater')
+
+    np.testing.assert_array_less(0.01, pval_low)
```
LacombeLouis, line 297: What is the probability of this breaking?

mapie/tests/test_regression.py
```diff
@@ -463,7 +509,7 @@ def test_linear_data_confidence_interval(strategy: str) -> None:
     """
     mapie = MapieRegressor(**STRATEGIES[strategy])
     mapie.fit(X_toy, y_toy)
-    y_pred, y_pis = mapie.predict(X_toy, alpha=0.2)
+    y_pred, y_pis = mapie.predict(X_toy, alpha=0.5)
```
LacombeLouis, line 512: I'm confused about why we change the level of `alpha`.

mapie/tests/test_regression.py
```diff
@@ -517,19 +563,16 @@ def test_results_prefit_naive() -> None:
...
-    mapie_reg = MapieRegressor(estimator=estimator, cv="prefit")
-    mapie_reg.fit(X_val, y_val)
-    _, y_pis = mapie_reg.predict(X_test, alpha=0.05)
+    mapie_reg = MapieRegressor(estimator=estimator, method="base", cv="prefit")
+    mapie_reg.fit(X_calib, y_calib)
+    _, y_pis = mapie_reg.predict(X_calib, alpha=0.05)
     width_mean = (y_pis[:, 1, 0] - y_pis[:, 0, 0]).mean()
     coverage = regression_coverage_score(
-        y_test, y_pis[:, 0, 0], y_pis[:, 1, 0]
+        y_calib, y_pis[:, 0, 0], y_pis[:, 1, 0]
```
LacombeLouis, line 575: Why don't we check on the test set?

LacombeLouis (changes requested):
Hey @thibaultcordier,
thank you for this PR. I think this should fix the issue that has been identified in the video. 

I've added some comments and I think to conclude the PR it would be beneficial to add a new issue to be able to setup any quantile method. 
Thank you

mapie/regression/regression.py
thibaultcordier: This is because the `split` method retains `nan` values for the data used to train the model. Because we need to keep the size of calibration dataset, we only sum those that have a defined conformity score.

mapie/tests/test_conformity_scores.py
thibaultcordier: Good question!

mapie/tests/test_conformity_scores.py
thibaultcordier: Because we don't have enough data for fit and calibrate methods (because of the change in the quantile formula). Because here we are simply checking the shape of the output, we don't care which method we use.

mapie/conformity_scores/conformity_scores.py
thibaultcordier: Good point, I will test that 👍

mapie/tests/test_conformity_scores.py
thibaultcordier: I have changed it to `NDArray`

mapie/tests/test_regression.py
thibaultcordier: I propose to run a FWER control procedure (Bonferroni correction) with p_fwer = 1% (see next update).

mapie/tests/test_regression.py
thibaultcordier: Because of the UPD of the quantile formula.
- we have a toy dataset with 6 exemples
- we divide it into two data sets with split method (50%) thus n_calib=3
- when we use alpha=0.2, we need at least 1/(0.2)=5 calibration data (but here we have 3)
- I propose to use alpha=0.5, we need at least 1/(0.5)=2 calibration data

mapie/tests/test_regression.py
thibaultcordier: Because we alerady do that in `test_linear_regression_results` for other strategies.
We can also see that we have obtained the same result as the `split` method in the same configuration.
WIDTH: "split": 3.89, "prefit": 3.89,
COVERAGE: "split": 0.956, "prefit": 0.956
Finally, as we are calculating the coverage on the calibration set, we can be sure that the empirical coverage is greater than the target coverage.

mapie/tests/test_conformity_scores.py
LacombeLouis: Can't it be a `float` or `NDArray`?

thibaultcordier: On line 385, we only have an array of floats. This is just in the context of the test, so it's not a problem for MAPIE's main content.

LacombeLouis (approved):
When the final comments are done, then I think we can finish this issue and make a release!

PR #463, opened 2024-06-11
FIX bug #462: Decoupling infinite interval production and quantile calculation

### LINKED ISSUE #462: Coverage validity not verified in MapieRegressor when producing infinite intervals
**Describe the bug**
In `MapieRegressor`, the quantile calculation does not allow us to check the coverage validity when producing infinite intervals.

**To Reproduce**
See #452.

**Expected behavior**
Decoupling infinite interval production and quantile calculation (not interference).

**Additional context**
See #452.

thibaultcordier:
# Description

Decoupling infinite interval production and quantile calculation

Fixes #462 

## Type of change

# How Has This Been Tested?

- [x] Test when alpha and the number of effective calibration samples are compatible
- [x] Test that the production of infinite intervals and the calculation of quantiles are properly decoupled

# Checklist


examples/regression/2-advanced-analysis/plot_coverage_validity.py
```diff
@@ -0,0 +1,528 @@
...
+
+##############################################################################
+# It can be seen that the two curves overlap, proving that both methods
+# produce the same results. Their effective coverage stabilizes between
+# the theoretical limits, always above the target coverage and converges
+# towards the exact coverage (i.e. expected according to the theory).
+
+
+##############################################################################
+# Experiment 2: Again but without fixing random_state
```
LacombeLouis, line 246: I would only keep the experiment 2 where you fix the `random_state`.

mapie/utils.py
```diff
@@ -412,6 +412,29 @@ def check_gamma(
         )
 
 
+def get_effective_calibration_samples(scores: NDArray, sym: bool):
```
LacombeLouis, line 415: Is the `sym` not by default `True`?

mapie/conformity_scores/conformity_scores.py
```diff
@@ -369,6 +384,11 @@ def get_bounds(
 
             By default ``False``.
 
+        allow_infinite_bounds: bool
```
LacombeLouis, line 387: Why do we call it `allow_infinite_bounds` and the `unbounded`?

examples/regression/2-advanced-analysis/plot_coverage_validity.py
```diff
@@ -0,0 +1,528 @@
...
+plt.title(r'$|D_{cal}| = $' + str(n_calib) + r' and $\delta = $' + str(delta))
+
+plt.legend(loc="upper right", ncol=2)
+plt.ylim(0.7, 1)
+plt.tight_layout()
+plt.show()
+
+
+##############################################################################
+# Experiment 4: Extensive experimentation on different delta and n_calib
```
LacombeLouis, line 433: I would only show the `reference`, `prefit`.

examples/regression/2-advanced-analysis/plot_coverage_validity.py
```diff
@@ -0,0 +1,528 @@
...
+plt.title(r'$|D_{cal}| = $' + str(n_calib) + r' and $\delta = $' + str(delta))
+
+plt.legend(loc="upper right", ncol=2)
+plt.ylim(0.7, 1)
+plt.tight_layout()
+plt.show()
+
+
+##############################################################################
+# Section 2: Comparison with different MAPIE CP methods
```
LacombeLouis, line 295: Not sure that this example adds a lot of insights. I would make this example shorter by getting rid of the `prefit_asym` part.

LacombeLouis (approved):
Hey @thibaultcordier,
I think the PR is great! Thank you for including the examples! This PR could be approved as is, but there are some small changes that I think could help make it clearer. Especially in the example, I don't think it's needed to add the `prefit_asym` part.

Thank you!

mapie/utils.py
thibaultcordier: The `sym` is by default `True` in `ConformityScore`. In any case, we need to compute the effective calibration sample size to compute the quantile.

mapie/conformity_scores/conformity_scores.py
thibaultcordier: In `MapieRegressor`, it is helpful to keep `allow_infinite_bounds` to understand what it is. In `get_bounds`, it is not helpful to understand the code because too long.

examples/regression/2-advanced-analysis/plot_coverage_validity.py
thibaultcordier: It is important to show the asymmetric case to avoid any misunderstanding/publication of an issue that is not an issue.

examples/regression/2-advanced-analysis/plot_coverage_validity.py
thibaultcordier: In practice, this part could be improved by adding more MAPIE methods (like split, CV+, CQR etc). In any case, we let the user test it themselves.

examples/regression/2-advanced-analysis/plot_coverage_validity.py
thibaultcordier: I have to keep this part, because the reader will no longer understand why in the next plot, the different methods no longer overlap when they did before.

Issue #231, opened 2022-11-02
Automated estimation of number resamplings given the size of the train data


ivan-marroquin:
**Is your feature request related to a problem? Please describe.**
When defining the "cv" splitter using the Subsample class, it is required to provide the "n_resamplings" and "n_samples". If the "n_resamplings" is not properly selected, the following warning message is raised:

"WARNING: at least one point of training set belongs to every resamplings. Increase the number of resamplings" 

**Describe the solution you'd like**
I think it will be beneficial if there is an automated way to estimate "n_resamplings" given the "n_samples". For instance, a user would choose to fix the "n_samples" in the following manner: n_samples= int(0.25 * gral_train_inputs.shape[0])

Then, the "n_resamplings" is determined accordingly to the size of the training data. 

**Describe alternatives you've considered**
In my case, I decided to fix the "n_samples" as shown above. But now, I have to do trail/error to find the minimum "n_resamplings" to avoid the warning message to ensure good statistical results.

Kind regards,
Ivan

vincentblot28:
Hi @ivan-marroquin, thank you for your issue. Indeed, automatically estimated the number of resampling is a good idea. Do you have a theoretical formula to do it ? If so feel free to share it with us, or even implement it into MAPIE a do a Pull Request

BaptisteCalot:
Hello everyone ! I tried to reproduce the bug without success. How am I supposed to get the warning message raised ? @vincentblot28

jawadhussein462:
Hello , 

I am closing this issue as the PR #238 resolves  it.

PR #464, opened 2024-06-12
238 giving a fraction of samples instead of a number of samples in the subsample class

BaptisteCalot:
# Description

Fix addressing issue 238. We introduce the ability to create a training set using the split method of the subsample class with a fraction between 0 and 1 representing the proportion of the training set. The option to specify an integer representing the number of elements in the training set is still retained

## Type of change

- Bug fix 

In the case where the attribute self.n_samples is a float, the feature n_samples used in the split method of Subsample class becomes self.n_samples * X.shape[0], taking its floor integer part

# How Has This Been Tested?

- Test test_split_SubSample_n_samples

We check that the training and test sets build using the split method are as expected with the given seed. We use two instances of Subsample, one with an integer n_samples and the other with an n_samples less than 1

# Checklist


HISTORY.rst
```diff
@@ -5,6 +5,8 @@ History
 0.8.5 (2024-**-**)
```
LacombeLouis, line 5: Note that this should be for 0.8.6.

mapie/subsample.py
```diff
@@ -74,9 +75,12 @@ def split(
             The testing set indices for that split.
         """
         indices = np.arange(_num_samples(X))
-        n_samples = (
-            self.n_samples if self.n_samples is not None else len(indices)
-        )
+        if self.n_samples is None:
+            n_samples = len(indices)
+        elif isinstance(self.n_samples, float) and 0 < self.n_samples < 1:
```
LacombeLouis, line 80: I think this should raise an error no? Say it's 1.2?

LacombeLouis (changes requested):
Thank you. I think we have some issues we want to clear up! But otherwise looks like a good PR. Note to also give a response to the issue #231!

mapie/subsample.py
```diff
@@ -22,9 +22,10 @@ class Subsample(BaseCrossValidator):
     ----------
     n_resamplings : int
         Number of resamplings. By default ``30``.
-    n_samples: int
+    n_samples: float
```
thibaultcordier, line 25:
```suggestion
    n_samples: Union[int, float]
```

mapie/subsample.py
thibaultcordier: I agree with this comment. Create a `check_n_samples_ratio` function in utils.py and create unit tests associated.

mapie/tests/test_subsample.py
```diff
@@ -32,6 +32,31 @@ def test_split_SubSample() -> None:
     np.testing.assert_equal(tests, tests_expected)
 
 
+def test_split_SubSample_n_samples() -> None:
```
thibaultcordier, line 35:
```suggestion
@pytest.mark.parametrize("n_samples", [0.8, 6])
def test_split_SubSample_n_samples(n_samples: Union[int, float]) -> None:
```
Prefer to use this convention when testing several values of a function.

thibaultcordier: However, I like your test but it has the wrong name. Suggest a test called `test_n_samples_equivalence' which tests that using `n_samples` and `n_ratio=n_samples/n_all` produces the same split.

thibaultcordier (changes requested):
Thank you for your PR. I suggest you write separate unit tests for better code maintenance.

mapie/subsample.py
```diff
@@ -74,9 +76,14 @@ def split(
...
         """
         indices = np.arange(_num_samples(X))
-        n_samples = (
-            self.n_samples if self.n_samples is not None else len(indices)
-        )
+        if self.n_samples is None:
+            n_samples = len(indices)
+        else:
+            n_samples = check_n_samples(X, self.n_samples)
+        # elif isinstance(self.n_samples, float) and 0 < self.n_samples < 1:
```
LacombeLouis, line 83: I don't think this is needed anymore?

mapie/subsample.py
```diff
@@ -74,9 +76,14 @@ def split(
             The testing set indices for that split.
         """
         indices = np.arange(_num_samples(X))
-        n_samples = (
-            self.n_samples if self.n_samples is not None else len(indices)
-        )
+        if self.n_samples is None:
```
LacombeLouis, line 79: Why isn't this included in `check_n_samples`?

mapie/tests/test_utils.py
```diff
@@ -508,3 +509,18 @@ def test_check_no_agg_cv_value_error(cv: Any) -> None:
         match=r"Allowed values must have the `get_n_splits` method"
     ):
         check_no_agg_cv(X_toy, cv, array)
+
+
+@pytest.mark.parametrize("X", [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
```
LacombeLouis, line 514: No need to test for so many values. Keep it simpler in this case. There is a choice to make between a bunch of tests and the time it takes to run them.

mapie/utils.py
```diff
@@ -1332,3 +1332,39 @@ def check_arrays_length(*arrays: NDArray) -> None:
...
+        n_samples
+
+    Raises
+    ------
+    ValueError
+        If n_samples is not an int or a float between 0 and 1.
+    """
+    if isinstance(n_samples, float) and 0 < n_samples < 1:
+        n_samples = int(np.floor(n_samples * X.shape[0]))
+    elif isinstance(n_samples, float) and n_samples > 1:
```
LacombeLouis, line 1364: What if `n_samples=-0.5`? Why are we only checking for `n_samples` greater than 1?

mapie/utils.py
```diff
@@ -1332,3 +1332,39 @@ def check_arrays_length(*arrays: NDArray) -> None:
         raise ValueError(
                 "There are arrays with different length"
             )
+
+
+def check_n_samples(
```
LacombeLouis, line 1337: What happens if `n_samples` is exactly `0` or `1`?

LacombeLouis (changes requested):

LacombeLouis (changes requested):
Hey @BaptisteCalot,
Thank you for this PR! I think there are still some adjustments to make! As mentioned in issue #231, I think it's important to have a default value.
Thank you!

mapie/tests/test_subsample.py
```diff
@@ -32,6 +32,31 @@ def test_split_SubSample() -> None:
...
+    test1 = np.concatenate([x[1] for x in cv1.split(X)])
+    train2 = np.concatenate([x[0] for x in cv2.split(X)])
+    test2 = np.concatenate([x[1] for x in cv2.split(X)])
+    train1_expected = np.array([2, 8, 4, 9, 1, 6, 7,
+                                3, 3, 5, 1, 2, 9, 8, 0, 6])
+    test1_expected = np.array([0, 5, 4, 7])
+    train2_expected = np.array([2, 8, 4, 9, 1, 6, 3, 5, 1, 2, 9, 8])
+    test2_expected = np.array([0, 3, 5, 7, 0, 4, 6, 7])
+    expected_n_samples_cv1 = int(np.floor(0.8 * X.shape[0]))
+    assert len(train1) == 2 * expected_n_samples_cv1
```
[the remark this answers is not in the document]
BaptisteCalot: I have taken it into account

mapie/subsample.py
BaptisteCalot: It is now Optional[Union[int, float]]

mapie/tests/test_utils.py
```diff
@@ -508,3 +510,19 @@ def test_check_no_agg_cv_value_error(cv: Any) -> None:
         match=r"Allowed values must have the `get_n_splits` method"
     ):
         check_no_agg_cv(X_toy, cv, array)
+
+
+@pytest.mark.parametrize("n_samples", [-5.5, -4, 0, 1.2])
```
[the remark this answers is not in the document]
BaptisteCalot: Thank you for your insightful comment Louis, I have taken it into account

mapie/utils.py
```diff
@@ -1355,3 +1355,50 @@ def check_arrays_length(*arrays: NDArray) -> None:
...
+                "are float in the range (0.0, 1.0) or"
+                " int in the range [1, inf)"
+                )
+    elif isinstance(n_samples, int) and n_samples <= 0:
+        raise ValueError(
+                "Invalid n_samples. Allowed values "
+                "are float in the range (0.0, 1.0) or"
+                " int in the range [1, inf)"
+            )
+    return int(n_samples)
```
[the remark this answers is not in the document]
LacombeLouis: So, when `replace=False`, what's the error?

LacombeLouis (approved):
Hey @BaptisteCalot,
Perfect, thank you! Please quickly mention the error that will be obtained when `replace=False`. Otherwise, I think this can be validated today!

mapie/tests/test_utils.py
```diff
@@ -508,3 +509,67 @@ def test_check_no_agg_cv_value_error(cv: Any) -> None:
...
+        match=re.escape(
+            r"Invalid n_samples. Allowed values "
+            r"are float in the range (0.0, 1.0) or"
+            r" int in the range [1, inf)"
+        )
+    ):
+        check_n_samples(X=X, n_samples=n_samples, indices=indices)
+
+
+@pytest.mark.parametrize("n_samples", [1.2, 2.5, 3.4])
```
LacombeLouis, line 562: You can put this test with the `test_invalid_n_samples_float_negative`. Good to separate, but this is a bit too much 😅. (So simply get rid of this test and add [-5.5, -4.3, -0.2, 1.2, 3.64]

mapie/utils.py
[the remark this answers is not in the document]
LacombeLouis: So, when `replace=False`, what's the error?

mapie/tests/test_utils.py
BaptisteCalot: Okay, I have just taken it into account

mapie/utils.py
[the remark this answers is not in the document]
BaptisteCalot: When replace = False, an error is raised by the resample function from scikit-learn that we use in the split method of Subsample : [link](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_indexing.py#L515)

mapie/utils.py
```diff
@@ -1355,3 +1355,56 @@ def check_arrays_length(*arrays: NDArray) -> None:
...
+                "Invalid n_samples. Allowed values "
+                "are float in the range (0.0, 1.0) or"
+                " int in the range [1, inf)"
+                )
+    elif isinstance(n_samples, int) and n_samples <= 0:
+        raise ValueError(
+                "Invalid n_samples. Allowed values "
+                "are float in the range (0.0, 1.0) or"
+                " int in the range [1, inf)"
+            )
```
thibaultcordier, line 1409:
```suggestion
        raise ValueError(
             "Invalid n_samples. Allowed values "
             "are float in the range (0.0, 1.0) or"
             " int in the range [1, inf)"
             )
```

thibaultcordier (approved):

PR #476, opened 2024-07-01
ENH: externalize MapieClassifier methods into several non-conformity scores

### LINKED ISSUE #467: Extending the NCS class to the classification task
Extending the NCS class to the classification task

Implement all existing classification NCS based on the `ConformityScore` interface.

thibaultcordier:
# Description

Outsource the `MapieClassifier` methods (LAC, APS, RAPS, TOPK) by extending the non-conformity score class to the classification task.

Fixes #467 

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

# How Has This Been Tested?

- [x] Test that existing code continues to work
- [x] Test that the proposed solution can enable new non-conformity scores

# Checklist


vincentblot28:
What a tour de force @thibaultcordier  !! Great improvement to the classification API ! 
 General comment about the example: as in the future we prefer to use directly the conformity scores (and the argument "method" might be removed) I think examples should follow the new way of using MapieClassifier

mapie/classification.py
```diff
@@ -149,9 +146,6 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):
     conformity_scores_: ArrayLike of shape (n_samples_train)
```
vincentblot28, line 146: I think that `conformity_scores_`has been removed to

vincentblot28: Actually, as this won't be a major release, we should have as few breaking changes as possible. So maybe we should keep those two attributes with a warning saying that users should prefer `mapie_cls.conformity_score.quantiles` to "mapie_cls.quantiles"

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,147 @@
...
+from mapie.conformity_scores.sets.utils import get_true_label_cumsum_proba
+from mapie.estimator.classifier import EnsembleClassifier
+
+from mapie._typing import NDArray
+from mapie.utils import compute_quantiles
+
+
+class APS(Naive):
+    """
+    Adaptive Prediction Sets (APS) method-based non-conformity score.
```
vincentblot28, line 16: outdatted docstring

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,147 @@
+from typing import Optional, cast
+
+import numpy as np
+from sklearn.dummy import check_random_state
+
+from mapie.conformity_scores.sets.naive import Naive
+from mapie.conformity_scores.sets.utils import get_true_label_cumsum_proba
```
vincentblot28, line 7: can we add `get_true_label_cumsum_proba` to the class method ? because it's only used for APS and RAPS. So if it's class method of APS, as RAPS inherits from APS, it will also be a method of RAPS

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,506 @@
...
+    quantiles_: ArrayLike of shape (n_alpha)
+        The quantiles estimated from ``get_sets`` method.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_external_attributes(
+        self,
+        method: str = 'naive',
```
vincentblot28, line 46: method is outdatted

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,506 @@
...
+        """
+        y_pred_proba = estimator.predict(X, agg_scores)
+        y_pred_proba = check_proba_normalized(y_pred_proba, axis=1)
+        if agg_scores != "crossval":
+            y_pred_proba = np.repeat(
+                y_pred_proba[:, :, np.newaxis], len(alpha_np), axis=2
+            )
+        return y_pred_proba
+
+    def get_conformity_quantiles(
```
vincentblot28, line 143:
```suggestion
    def get_cs_quantiles(
```

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,506 @@
...
+        y_proba_last_cumsumed = np.squeeze(
+            np.take_along_axis(
+                y_pred_proba_cumsum,
+                y_pred_index_last,
+                axis=1
+            ), axis=1
+        )
+
+        # get the V parameter from Romano+(2020) or Angelopoulos+(2020)
+        vs = self._compute_vs_parameter(
```
vincentblot28, line 388: there is no vs need in the naive method

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,506 @@
...
+        )
+        return prediction_sets
+
+    def get_prediction_sets(
+        self,
+        y_pred_proba: NDArray,
+        conformity_scores: NDArray,
+        alpha_np: NDArray,
+        estimator: EnsembleClassifier,
+        agg_scores: Optional[str] = "mean",
```
vincentblot28, line 415: Can you includ agg_scores and include_last_label to the kwargs as none of them is used for naive conformity score ?

mapie/conformity_scores/sets/raps.py
```diff
@@ -0,0 +1,477 @@
...
+    quantiles_: ArrayLike of shape (n_alpha)
+        The quantiles estimated from ``get_sets`` method.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_external_attributes(
+        self,
+        method: str = 'raps',
```
vincentblot28, line 65: outdatted

mapie/classification.py
thibaultcordier: We should instead add `conformity_score_function_` as a new attribute.

thibaultcordier: At the same time, I can leave `quantiles_` accessible as before.

mapie/conformity_scores/sets/naive.py
thibaultcordier: vs is needed when `include_last_label == "randomized"`, do we have a condition that prohibits `method='naive'` and `include_last_label='randomized'`?

mapie/classification.py
thibaultcordier: Done 👍

mapie/conformity_scores/sets/naive.py
vincentblot28: yes we should, acutually naive method should always have include_last_label=True

thibaultcordier: This condition is not currently verified, is that correct? Should this therefore be the subject of another PR?

mapie/conformity_scores/sets/raps.py
thibaultcordier: Done 👍

mapie/conformity_scores/sets/naive.py
thibaultcordier: Done 👍

mapie/conformity_scores/sets/aps.py
thibaultcordier: Done 👍

mapie/conformity_scores/sets/naive.py
thibaultcordier: Done 👍

mapie/classification.py
vincentblot28: and to prevent any breaking changes, you can also add `conformity_scores_' as class attributes

mapie/conformity_scores/sets/naive.py
vincentblot28: I think we can add it in this PR, it will make the code easier to read

vincentblot28: @LacombeLouis do you agree ?

mapie/classification.py
thibaultcordier: It's already done, right?

mapie/conformity_scores/classification.py
LacombeLouis, line 1: Would it be more appropriate to have this in `mapie/conformity_scores/sets/base`? If not why?

mapie/conformity_scores/sets/raps.py
```diff
@@ -0,0 +1,427 @@
...
+
+            By default, "None" but must be set to work.
+
+        position_raps: NDArray of shape (n_samples,)
+            Position of the points in the split set for the RAPS method
+            (split data). These positions are returned by the function
+            ``get_true_label_position``.
+
+            By default, "None" but must be set to work.
+
```
LacombeLouis, line 283: Don't forget to add `**kwargs`. Note that there seems like there are multiple times where it might have been forgotten.

LacombeLouis: It's a question whether we want to add it everywhere or not. @vincentblot28 @thibaultcordier ?

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Conformity scores
+        conformity_scores, self.cutoff = (
+            get_true_label_cumsum_proba(y, y_pred, classes)
+        )
+        y_proba_true = np.take_along_axis(
+            y_pred, y_enc.reshape(-1, 1), axis=1
+        )
+        random_state = check_random_state(self.random_state)
+        random_state = cast(np.random.RandomState, random_state)
```
LacombeLouis, line 79: Does the `check_random_state` not already do this?

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,127 @@
...
+        u = random_state.uniform(size=len(y_pred)).reshape(-1, 1)
+        conformity_scores -= u * y_proba_true
```
LacombeLouis, line 81: I would rather we use numpy to make multiplications.

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,127 @@
...
+        --------
+        NDArray
+            Array of quantiles with respect to alpha_np.
+        """
+        n = len(conformity_scores)
+
+        if estimator.cv == "prefit" or agg_scores in ["mean"]:
+            quantiles_ = compute_quantiles(conformity_scores, alpha_np)
+        else:
+            quantiles_ = (n + 1) * (1 - alpha_np)
```
LacombeLouis, line 125: Why don't we use compute quantiles?

mapie/tests/test_classification.py
```diff
@@ -1606,28 +1612,16 @@ def test_method_error_in_fit(monkeypatch: Any, method: str) -> None:
         mapie_clf.fit(X_toy, y_toy)
 
 
-@pytest.mark.parametrize("method", WRONG_METHODS)
```
LacombeLouis, line 1609: Why do we no longer perform this test?

mapie/tests/test_conformity_scores_sets.py
```diff
@@ -0,0 +1,37 @@
...
+import pytest
+
+# from mapie._typing import ArrayLike, NDArray
+from mapie.conformity_scores import BaseClassificationScore
+from mapie.conformity_scores.sets import APS, LAC, TopK
+from mapie.conformity_scores.utils import check_classification_conformity_score
+
+
+cs_list = [None, LAC(), APS(), TopK()]
+method_list = [None, 'naive', 'aps', 'raps', 'lac', 'top_k']
```
LacombeLouis, line 12: Should we add wrong methods here? (and a test to go with it)

LacombeLouis (commented):
Hey @thibaultcordier,
Thank you for this PR. Extremely well-thought-out and very detailed. Some small questions here and there. Looking forward to have this new PR included!

mapie/conformity_scores/interface.py
```diff
@@ -0,0 +1,153 @@
...
+            Predicted target values.
+
+        Returns
+        -------
+        NDArray of shape (n_samples, ...)
+            Conformity scores.
+        """
+
+    @staticmethod
+    def get_quantile(
```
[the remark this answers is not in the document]
thibaultcordier: mapie/conformity_scores/utils.py

mapie/conformity_scores/sets/utils.py
```diff
@@ -0,0 +1,204 @@
...
+    index_sorted = np.fliplr(np.argsort(y_pred_proba, axis=1))
+    y_pred_sorted = np.take_along_axis(y_pred_proba, index_sorted, axis=1)
+    y_true_sorted = np.take_along_axis(y_true, index_sorted, axis=1)
+    y_pred_sorted_cumsum = np.cumsum(y_pred_sorted, axis=1)
+    cutoff = np.argmax(y_true_sorted, axis=1)
+    true_label_cumsum_proba = np.take_along_axis(
+        y_pred_sorted_cumsum, cutoff.reshape(-1, 1), axis=1
+    )
+
+    return true_label_cumsum_proba, cutoff + 1
```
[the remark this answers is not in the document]
thibaultcordier: It was like that before, I didn't touch anything

mapie/conformity_scores/sets/utils.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the sum of the scores is not equal to one.
+    """
+    sum_proba = np.sum(y_pred_proba, axis=axis)
+    err_msg = "The sum of the scores is not equal to one."
+    np.testing.assert_allclose(sum_proba, 1, err_msg=err_msg, rtol=1e-5)
+    y_pred_proba = cast(NDArray, y_pred_proba).astype(np.float64)
```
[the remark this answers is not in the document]
thibaultcordier: It was like that before, I didn't touch it during the refactoring

mapie/classification.py
```diff
@@ -1287,149 +726,26 @@ def predict(
...
+        prediction_sets = self.conformity_score_function_.predict_set(
+            X, alpha_np,
+            estimator=self.estimator_,
+            conformity_scores=self.conformity_scores_,
+            include_last_label=include_last_label,
+            agg_scores=agg_scores,
+            **kwargs
+        )
+
+        self.quantiles_ = self.conformity_score_function_.quantiles_
```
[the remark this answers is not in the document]
thibaultcordier: This was a request from @vincentblot28 to keep self.quantiles_ already existing for classification

mapie/conformity_scores/classification.py
thibaultcordier: Everything is possible, it doesn't change the logic of the code. I put this here to reduce the size of modules which were starting to have a lot of files.

mapie/conformity_scores/sets/raps.py
thibaultcordier: I use **kwargs to allow increasing the size of input parameters in child classes without the parent class writing a warning. In practice, the user does not need to know that **kwargs if using the target class.

thibaultcordier: It's just a developer trick, which does nothing for the end user except confusion. I prefer not to put that.

mapie/conformity_scores/sets/aps.py
thibaultcordier: It also manages the default parameters (if None or if integer)

mapie/conformity_scores/sets/aps.py
thibaultcordier: It was like that before, I didn't touch it during the refactoring

mapie/conformity_scores/sets/aps.py
thibaultcordier: It was like that before, I didn't touch it during the refactoring

mapie/tests/test_classification.py
thibaultcordier: This test tested, in the case where the user changes the method value between fit and predict, that an error was reported.

But now, we manage non-conformity scores. Due to the logic of the refacto, this test in the code no longer makes sense because we do not modify the non-confnormity score as we want.

1) We do not do this test for MapieRegressor
2) We do not do this test if the estimator has changed between the fit and the predict

mapie/classification.py
vincentblot28: yes indeed ! perfect

mapie/conformity_scores/regression.py
```diff
@@ -337,10 +246,10 @@ def _beta_optimize(
 
     def get_bounds(
         self,
-        X: ArrayLike,
+        X: NDArray,
```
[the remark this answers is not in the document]
thibaultcordier: Because of `make type-check`

mapie/conformity_scores/regression.py
```diff
@@ -3,79 +3,74 @@
...
         self.consistency_check = consistency_check
         self.eps = eps
 
     @abstractmethod
     def get_signed_conformity_scores(
         self,
-        X: ArrayLike,
-        y: ArrayLike,
-        y_pred: ArrayLike,
+        y: NDArray,
```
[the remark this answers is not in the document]
thibaultcordier: Because of `make type-check`

mapie/conformity_scores/bounds/absolute.py
```diff
@@ -0,0 +1,52 @@
...
+
+class AbsoluteConformityScore(BaseRegressionScore):
+    """
+    Absolute conformity score.
+
+    The signed conformity score = y - y_pred.
+    The conformity score is symmetrical.
+
+    This is appropriate when the confidence interval is symmetrical and
+    its range is approximatively the same over the range of predicted values.
```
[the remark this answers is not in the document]
thibaultcordier: This is not the point of this PR.

mapie/conformity_scores/bounds/gamma.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Gamma conformity score.
+
+    The signed conformity score = (y - y_pred) / y_pred.
+    The conformity score is not symmetrical.
+
+    This is appropriate when the confidence interval is not symmetrical and
+    its range depends on the predicted values. Like the Gamma distribution,
+    its support is limited to strictly positive reals.
+    """
```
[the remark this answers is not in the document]
thibaultcordier: This is not the point of this PR.

mapie/conformity_scores/sets/naive.py
vincentblot28: it's not linked

mapie/conformity_scores/bounds/residuals.py
```diff
@@ -433,6 +312,13 @@ def get_estimation_distribution(
         ``conformity_scores`` can be either the conformity scores or
         the quantile of the conformity scores.
         """
+        if X is None:
```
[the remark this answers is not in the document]
thibaultcordier: This is necessary only here which depends on the method signature which differs from that of the parent class.

mapie/_machine_precision.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
 
-EPSILON = np.finfo(np.float64).eps
+EPSILON = np.float64(1e-8)
```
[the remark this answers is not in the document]
thibaultcordier: Because too restrictive for regression and small enough to check equalities for regression and classification.

mapie/conformity_scores/bounds/absolute.py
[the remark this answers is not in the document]
LacombeLouis: Could we create an issue then?

mapie/conformity_scores/interface.py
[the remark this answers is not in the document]
LacombeLouis: Unclear how this answers my question. To me, it would seem like it is used for both?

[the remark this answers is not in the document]
thibaultcordier: Wrong comment, sorry!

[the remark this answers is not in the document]
thibaultcordier: In the medium term, we would like this to be the case. But this will require a specific PR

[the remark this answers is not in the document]
thibaultcordier: In the short term, the classification uses `compute_quantile`

mapie/conformity_scores/regression.py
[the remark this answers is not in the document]
LacombeLouis: Fair, we do make the casting before using these functions.

mapie/conformity_scores/sets/utils.py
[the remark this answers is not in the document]
LacombeLouis: Could we improve it here? Not a big difference, just code readability.

mapie/conformity_scores/checks.py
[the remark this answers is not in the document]
thibaultcordier: mapie/conformity_scores/utils.py

mapie/conformity_scores/sets/utils.py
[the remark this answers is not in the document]
LacombeLouis: Then, we could create a separate issue, or solve it here.

mapie/conformity_scores/classification.py
LacombeLouis: I feel like it would make it more clear. It's up to debate, but it feels like in the setup, it would feel nicer like this. What do you think @vincentblot28?

mapie/conformity_scores/sets/aps.py
LacombeLouis: To me, it seems like it already returns a `np.random.RandomState`: [link](https://github.com/scikit-learn/scikit-learn/blob/70fdc843a4b8182d97a3508c1a426acc5e87e980/sklearn/utils/validation.py#L1411)

mapie/conformity_scores/sets/aps.py
LacombeLouis: I would perform the change, or create an issue.

mapie/conformity_scores/sets/aps.py
LacombeLouis: Just to check, @vincentblot28, is this the behavior expected?

mapie/conformity_scores/bounds/absolute.py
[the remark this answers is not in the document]
thibaultcordier: Done

mapie/conformity_scores/regression.py
```diff
@@ -160,123 +142,50 @@ def check_consistency(
...
-            Boolean specifying whether infinite prediction intervals
-            could be produced (when alpha_np is greater than or equal to 1.).
-
-            By default ``False``.
 
         Returns
         -------
-        NDArray of shape (1, n_alpha) or (n_samples, n_alpha)
-            The quantile of the conformity scores.
+        NDArray of shape (n_samples, ...)
```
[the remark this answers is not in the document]
thibaultcordier: Done

mapie/conformity_scores/sets/aps.py
thibaultcordier: And why do you want to use numpy here? It is not clear that the operation is element-wise?

mapie/conformity_scores/bounds/absolute.py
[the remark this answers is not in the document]
LacombeLouis: Thank you, #478.

mapie/conformity_scores/interface.py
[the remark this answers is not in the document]
LacombeLouis: Indeed, thank you. Could we create an issue for this? Thank you!

mapie/conformity_scores/sets/aps.py
LacombeLouis: Simply because it feels a bit convoluted. But if you don't agree, then it's okay for me. @vincentblot28

mapie/conformity_scores/interface.py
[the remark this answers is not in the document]
thibaultcordier: Done 👍

mapie/conformity_scores/sets/aps.py
thibaultcordier: Done

mapie/conformity_scores/sets/aps.py
thibaultcordier: We've talked it over with @vincentblot28 and it's normal for him, so there's no problem.

mapie/conformity_scores/sets/aps.py
thibaultcordier: We'll leave it at that 👍

mapie/classification.py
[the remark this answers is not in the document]
thibaultcordier: Done 👍

mapie/conformity_scores/sets/aps.py
thibaultcordier: We'll leave it at that 👍

mapie/conformity_scores/sets/utils.py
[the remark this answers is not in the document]
thibaultcordier: Done 👍

mapie/conformity_scores/sets/utils.py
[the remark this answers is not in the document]
thibaultcordier: Done 👍

mapie/conformity_scores/classification.py
thibaultcordier: I've tried to do this but I'm having a problem with the circular import. I'm giving up. If it's "really" a code structure that we want to impose, we'll change it. In practice, the user won't notice the difference.

thibaultcordier: We'll leave it at that 👍

mapie/conformity_scores/sets/naive.py
thibaultcordier: Done 👍 (move this method in APS class)

mapie/conformity_scores/sets/naive.py
thibaultcordier: Done 👍 (I have removed them because they are useless in the naive method)

mapie/conformity_scores/sets/aps.py
thibaultcordier: Done 👍

mapie/tests/test_conformity_scores_sets.py
thibaultcordier: Done 👍

mapie/conformity_scores/interface.py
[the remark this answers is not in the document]
LacombeLouis: Issue reference: #479

LacombeLouis (approved):

mapie/classification.py
```diff
@@ -47,7 +44,7 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):
         Method to choose for prediction interval estimates.
         Choose among:
 
-        - ``"naive"``, sum of the probabilities until the 1-alpha thresold.
+        - ``"naive"``, sum of the probabilities until the 1-alpha threshold.
```
vincentblot28, line 47: conformity_score not in the docstring

mapie/conformity_scores/utils.py
```diff
@@ -1,26 +1,96 @@
...
+    ... except Exception as exception:
+    ...     print(exception)
+    ...
+    Invalid conformity_score argument.
+    Must be None or a ConformityScore instance.
+    """
+    allowed_methods = ['lac', 'naive', 'aps', 'raps', 'top_k']
+    deprecated_methods = ['score', 'cumulated_score']
+    if method is not None:
+        if method in ['score', 'lac']:
```
vincentblot28, line 73: add depreciation warning to say that the user should use the conformity_score argument instead of method

vincentblot28: maybe it should be added in the `check_depreciated` method of MapieClassifier

mapie/classification.py
```diff
@@ -208,6 +209,7 @@ def __init__(
         cv: Optional[Union[int, str, BaseCrossValidator]] = None,
         test_size: Optional[Union[int, float]] = None,
         n_jobs: Optional[int] = None,
+        conformity_score: Optional[BaseClassificationScore] = None,
```
vincentblot28, line 212: default conformity_score should be LAC() and defautl method should be None in order to get used to define the conformity_score and not the method

vincentblot28: We should also change their order

mapie/classification.py
```diff
@@ -1109,6 +581,7 @@ def fit(
         """
         # Checks
         (estimator,
+         self.conformity_score_function_,
```
vincentblot28, line 584: Just a few lines bellow you check `if self.method == "raps"` but if we use the conformity score RAPS() then self.method will be None. In this case this piece of code won't work if `conformity_score=RAPS()` (maybe a test is missing)

mapie/classification.py
```diff
@@ -1158,35 +631,10 @@ def fit(
...
-            # Here we reorder the labels by decreasing probability
-            # and get the position of each label from decreasing
-            # probability
-            self.conformity_scores_ = get_true_label_position(
-                y_pred_proba, y_enc
-            )
-        else:
-            raise ValueError(
-                "Invalid method. " f"Allowed values are {self.valid_methods_}."
+        # Compute the conformity scores
```
vincentblot28, line 634: why don't we put those if conditions directly into the conformity scores ? Then all the code will be simplified as their won't be any need to have kwargs for the `predict_set` method

mapie/classification.py
```diff
@@ -1287,149 +726,26 @@ def predict(
...
-        if agg_scores != "crossval":
-            y_pred_proba = np.repeat(
-                y_pred_proba[:, :, np.newaxis], len(alpha_np), axis=2
-            )
-
-        # Choice of the quantile
-        if self.method == "naive":
-            self.quantiles_ = 1 - alpha_np
+        # Estimate prediction sets
+        if self.method == "raps":
```
vincentblot28, line 730: Same remark as above : this could be included directly in the non conformity score if we perform the raps split their

mapie/conformity_scores/sets/lac.py
```diff
@@ -0,0 +1,252 @@
...
+        Pseudo random number generator state.
+
+    quantiles_: ArrayLike of shape (n_alpha)
+        The quantiles estimated from ``get_sets`` method.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_external_attributes(
```
vincentblot28, line 43: is it necessary to re-define this method in the child class ?

mapie/conformity_scores/sets/topk.py
```diff
@@ -0,0 +1,227 @@
...
+        Pseudo random number generator state.
+
+    quantiles_: ArrayLike of shape (n_alpha)
+        The quantiles estimated from ``get_sets`` method.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_external_attributes(
```
vincentblot28, line 47: same as for LAC : is it needed ?

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,300 @@
...
+        Pseudo random number generator state.
+
+    quantiles_: ArrayLike of shape (n_alpha)
+        The quantiles estimated from ``get_sets`` method.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def set_external_attributes(
```
vincentblot28, line 35: same as for LAC and TopK

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,300 @@
...
+        )
+        # sort probabilities by decreasing order
+        y_pred_proba_sorted = np.take_along_axis(
+            y_pred_proba, index_sorted, axis=1
+        )
+        # get sorted cumulated score
+        y_pred_proba_sorted_cumsum = np.cumsum(y_pred_proba_sorted, axis=1)
+        y_pred_proba_sorted_cumsum = self._add_regualization(
+            y_pred_proba_sorted_cumsum, **kwargs
+        )
```
vincentblot28, line 216:
```suggestion
        y_pred_proba_sorted_cumsum = self._add_regualization(
            y_pred_proba_sorted_cumsum, **kwargs
        )  # Do nothing as no regularization for the naive method
```

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,438 @@
...
+        Pseudo random number generator state.
+
+    quantiles_: ArrayLike of shape (n_alpha)
+        The quantiles estimated from ``get_sets`` method.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    def get_predictions(
```
vincentblot28, line 45: where is this method used ?

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,438 @@
...
+            Threshold to compare with y_proba_last_cumsum, can be either:
+
+            - the quantiles associated with alpha values when
+                ``cv`` == "prefit", ``cv`` == "split"
+                or ``agg_scores`` is "mean"
+
+            - the conformity score from training samples otherwise (i.e., when
+            ``cv`` is CV splitter and ``agg_scores`` is "crossval")
+
+        method: str
```
vincentblot28, line 291: outdatted docstring

mapie/classification.py
thibaultcordier: Thanks for the reminder! Done 👍

mapie/conformity_scores/sets/raps.py
```diff
@@ -0,0 +1,426 @@
...
+    ) -> Union[NDArray, float]:
+        """
+        Find the optimal value of lambda for each alpha.
+
+        Parameters
+        ----------
+        y_pred_proba_raps: NDArray of shape (n_samples, n_labels, n_alphas)
+            Predictions of the model repeated on the last axis as many times
+            as the number of alphas
+
```
vincentblot28, line 158: Outdated docstraing

mapie/classification.py
thibaultcordier: To ensure that notebooks do not break, we cannot change the order (in the case where the user instantiates an object with the attributes but without mentioning the keys).

thibaultcordier: Thank you for the first comment, which I will take into account.

mapie/tests/test_classification.py
```diff
@@ -23,6 +23,8 @@
 
 from mapie._typing import ArrayLike, NDArray
 from mapie.classification import MapieClassifier
+from mapie.conformity_scores import APS, RAPS
+from mapie.conformity_scores.sets.utils import check_proba_normalized
```
vincentblot28, line 27: We must test that we have exacltly the same result when method is specified and when conformity score is specified (for all conformity scores)

vincentblot28: I'm also wondering if for future the future, we should write all our tests with conformity_score rather than with method

mapie/classification.py
thibaultcordier: I'm having issues with sklearn's `check_estimator.`
- I had to do like MapieRegressor for which `conformity_score=None` by default.
- This doesn't change `conformity_score` to take precedence over `method`.
- This time, `method=None` by default.

The consequence is that I had to delete a test which was no longer useful with more default values ​​other than None.

thibaultcordier: Done 👍

mapie/conformity_scores/utils.py
thibaultcordier: I moved it in utils.py, but it is still possible to do that definitively.

mapie/conformity_scores/utils.py
```diff
@@ -1,26 +1,163 @@
...
+    ...     check_classification_conformity_score
+    ... )
+    >>> try:
+    ...     check_classification_conformity_score(1)
+    ... except Exception as exception:
+    ...     print(exception)
+    ...
+    Invalid conformity_score argument.
+    Must be None or a ConformityScore instance.
+    """
```
vincentblot28, line 137: Maybe cleaner if you create a dictionnary to map method to conformity scores

mapie/conformity_scores/utils.py
vincentblot28: Yes it would be great to have a depreciation warning if method is not None

mapie/conformity_scores/utils.py
vincentblot28: Moreover, as allowed methods are also defined in MapieClassifier, the values for `allowed_methods` should be an argument of this function

mapie/tests/test_classification.py
```diff
@@ -1504,7 +1500,8 @@ def test_cumulated_scores() -> None:
         include_last_label=True,
         alpha=alpha
     )
-    np.testing.assert_allclose(mapie_clf.quantiles_, quantile)
+    computed_quantile = mapie_clf.conformity_score_function_.quantiles_
```
vincentblot28, line 1503: this change is now longer required

mapie/tests/test_classification.py
```diff
@@ -1532,7 +1529,8 @@ def test_image_cumulated_scores(X: Dict[str, ArrayLike]) -> None:
         include_last_label=True,
         alpha=alpha
     )
-    np.testing.assert_allclose(mapie.quantiles_, quantile)
+    computed_quantile = mapie.conformity_score_function_.quantiles_
```
vincentblot28, line 1532: change no longer required

mapie/tests/test_classification.py
```diff
@@ -1744,12 +1729,9 @@ def test_regularize_conf_scores_shape(k_lambda) -> None:
     Test that the conformity scores have the correct shape.
     """
     lambda_, k = k_lambda[0], k_lambda[1]
-    args_init, _ = STRATEGIES["raps"]
-    clf = LogisticRegression().fit(X, y)
-    mapie_clf = MapieClassifier(estimator=clf, **args_init)
     conf_scores = np.random.rand(100, 1)
     cutoff = np.cumsum(np.ones(conf_scores.shape)) - 1
-    reg_conf_scores = mapie_clf._regularize_conformity_score(
+    reg_conf_scores = RAPS._regularize_conformity_score(
```
vincentblot28, line 1734: this now test the conformity score and not the MapieClassifier class, so this test should go in another file

mapie/tests/test_classification.py
```diff
@@ -1768,9 +1750,8 @@ def test_get_true_label_cumsum_proba_shape() -> None:
         estimator=clf, random_state=random_state
     )
     mapie_clf.fit(X, y)
-    cumsum_proba, cutoff = mapie_clf._get_true_label_cumsum_proba(
-        y, y_pred
-    )
+    classes = mapie_clf.classes_
```
vincentblot28, line 1753: Same : this test should go in an other file

mapie/tests/test_classification.py
```diff
@@ -1787,8 +1768,9 @@ def test_get_true_label_cumsum_proba_result() -> None:
         estimator=clf, random_state=random_state
     )
     mapie_clf.fit(X_toy, y_toy)
-    cumsum_proba, cutoff = mapie_clf._get_true_label_cumsum_proba(
-        y_toy, y_pred
+    classes = mapie_clf.classes_
```
vincentblot28, line 1771: This test should go in an other file

mapie/conformity_scores/utils.py
thibaultcordier: Done 👍 everything (like `allowed_methods`) is managed directly by the check function

mapie/conformity_scores/utils.py
thibaultcordier: Done 👍

mapie/tests/test_classification.py
thibaultcordier: Done 👍

mapie/tests/test_classification.py
thibaultcordier: Done 👍

mapie/tests/test_classification.py
thibaultcordier: Done 👍

mapie/tests/test_classification.py
thibaultcordier: Done 👍

mapie/tests/test_classification.py
thibaultcordier: Done 👍

mapie/tests/test_classification.py
```diff
@@ -1827,12 +1809,12 @@ def test_get_last_included_proba_shape(k_lambda, strategy):
         y_pred_proba[:, :, np.newaxis], len(thresholds), axis=2
     )
 
-    mapie = MapieClassifier(estimator=clf, **STRATEGIES[strategy][0])
     include_last_label = STRATEGIES[strategy][1]["include_last_label"]
-    y_p_p_c, y_p_i_l, y_p_p_i_l = mapie._get_last_included_proba(
-        y_pred_proba, thresholds,
-        include_last_label, lambda_, k
-    )
+    y_p_p_c, y_p_i_l, y_p_p_i_l = \
```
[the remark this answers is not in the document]
thibaultcordier: Done 👍

mapie/tests/test_conformity_scores.py
```diff
@@ -5,31 +5,34 @@
 from sklearn.preprocessing import PolynomialFeatures
 
 from mapie._typing import ArrayLike, NDArray
-from mapie.conformity_scores import (AbsoluteConformityScore, ConformityScore,
-                                     GammaConformityScore,
-                                     ResidualNormalisedScore)
+from mapie.conformity_scores import (
```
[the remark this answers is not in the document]
thibaultcordier: Done 👍

mapie/conformity_scores/sets/aps.py
thibaultcordier: Done 👍

mapie/conformity_scores/sets/naive.py
thibaultcordier: Done 👍

mapie/classification.py
thibaultcordier: Done 👍 I hope the proposal suits you

mapie/classification.py
thibaultcordier: Done 👍 I hope the proposal suits you

mapie/conformity_scores/sets/lac.py
thibaultcordier: it's fine to leave it for future changes

mapie/conformity_scores/sets/aps.py
thibaultcordier: mapie/conformity_scores/classification.py:l155 in `get_sets` method

mapie/conformity_scores/sets/naive.py
thibaultcordier: it's fine to leave it for future changes

mapie/classification.py
thibaultcordier: Tell me if the proposed solution solves this problem:
- everything that depends on the logic of the RAPS code is included in the eponymous class.
- everything that depends on checking that the correct values have been assigned is managed in the `utils.py` file in the `conformity_scores` module.

thibaultcordier: Done 👍

mapie/tests/test_classification.py
thibaultcordier: Done 👍 (see `test_same_result_with_score_and_method`)

mapie/classification.py
```diff
@@ -979,91 +367,44 @@ def _check_fit_parameter(
...
+            estimator, cs_estimator, cv, X, y, y_enc, sample_weight, groups
+        )
 
     def fit(
         self,
         X: ArrayLike,
         y: ArrayLike,
         sample_weight: Optional[ArrayLike] = None,
-        size_raps: Optional[float] = 0.2,
+        size_raps: Optional[float] = None,
```
vincentblot28, line 407: why did you change this value ?

thibaultcordier: To give priority to the RAPS class. If I forgot to put a deprecatedwarning, I have to add it

mapie/conformity_scores/sets/lac.py
vincentblot28: If in the future we need to redefine it, then we will rewrite it. However, I think to make the code more concise, we should remove it

mapie/conformity_scores/sets/naive.py
vincentblot28: If in the future we need to redefine it, then we will rewrite it. However, I think to make the code more concise, we should remove it.

mapie/conformity_scores/sets/topk.py
vincentblot28: If in the future we need to redefine it, then we will rewrite it. However, I think to make the code more concise, we should remove it.

mapie/conformity_scores/utils.py
```diff
@@ -1,26 +1,173 @@
...
-    NDArray of shape (n_samples, 1)
-        Position of the true label in the prediction.
+def check_classification_conformity_score(
+    conformity_score: Optional[BaseClassificationScore] = None,
+    method: Optional[str] = None,
+) -> BaseClassificationScore:
     """
-    index = np.argsort(np.fliplr(np.argsort(y_pred_proba, axis=1)))
-    position = np.take_along_axis(index, y.reshape(-1, 1), axis=1)
+    Check parameter ``conformity_score`` for classification task.
```
vincentblot28, line 130: Missing docstring for the parameters

mapie/conformity_scores/utils.py
```diff
@@ -1,26 +1,173 @@
...
+    if method is not None:
+        if isinstance(method, str) and method in METHOD_SCORE_MAP:
+            _check_depreciated(method)
+            return METHOD_SCORE_MAP[method]()
+        else:
+            raise ValueError(
+                "Invalid method. "
+                f"Allowed values are {list(METHOD_SCORE_MAP.keys())}."
+            )
+    else:
```
vincentblot28, line 169: I think this else statement is at the wrong place as if it is not None and it's not an instance of BaseClassificaitonScore, then this raise is not accessed. Add a test to

mapie/conformity_scores/utils.py
thibaultcordier: Done 👍

mapie/conformity_scores/utils.py
thibaultcordier: The test already exists and check if it is None, then if it is an instance of str, but not a correct string, then raise an error (see `test_check_wrong_classification_method` l101 and `test_check_conflict_parameters` l79 of `mapie/tests/test_conformity_scores_sets.py`).

thibaultcordier: However, I forgot to raise an error when `conformity_score` is not of the right type, which I will add right away

mapie/conformity_scores/utils.py
```diff
@@ -1,26 +1,197 @@
...
-    return position
+    y: NDArray of shape (n_samples,)
+        Training labels.
+
+    Raises
+    ------
+    ValueError
+        If type of target is binary and method is not ``"lac"``
+        or ``"score"`` or if type of target is not multi-class.
+    """
```
vincentblot28, line 114: Why adding the lambda ? you can write `score: APS` so the method is not instanciate and you do it only when you call it

thibaultcordier: type-check problem

vincentblot28: Should be at the top of the file

mapie/conformity_scores/utils.py
vincentblot28: I think that the allowed methods should be handled in MapieClassifier

mapie/conformity_scores/utils.py
thibaultcordier: Done 👍

mapie/classification.py
vincentblot28: Open question : should this value be handled directly by the conformity score ?

mapie/conformity_scores/utils.py
thibaultcordier: No, I think it's better to keep a black box perspective. And it remains within the logic of outsourcing the scores with MapieClassifier.

mapie/classification.py
thibaultcordier: We would like MapieClassifier to no longer have a dependency on RAPS. Then yes, this value shoud be handled directly by the conformity score.

mapie/conformity_scores/utils.py
thibaultcordier: Done 👍

mapie/conformity_scores/utils.py
```diff
@@ -1,26 +1,199 @@
...
+            )
+        if isinstance(conformity_score, BaseClassificationScore):
+            return conformity_score
+        else:
+            raise ValueError(
+                "Invalid conformity_score argument.\n"
+                "Must be None or a BaseClassificationScore instance."
+            )
+    elif method is not None:
+        if isinstance(method, str) and method in METHOD_SCORE_MAP:
```
vincentblot28, line 190: no need to check that it's a string as if it's in `METHOD_SCORE_MAP` then it's a string

thibaultcordier: Done 👍

thibaultcordier: But doens't work if the value you test is a list (unshashable value)

mapie/regression/regression.py
```diff
@@ -639,16 +643,15 @@ def predict(
...
-                    X,
-                    self.estimator_,
-                    self.conformity_scores_,
-                    alpha_np,
-                    ensemble=ensemble,
-                    method=self.method,
-                    optimize_beta=optimize_beta,
-                    allow_infinite_bounds=allow_infinite_bounds
-                )
+            outputs = self.conformity_score_function_.predict_set(
```
candicemyt, line 646: predict_set function is not required, why not just call get_bounds here ?

mapie/classification.py
```diff
@@ -1287,149 +584,15 @@ def predict(
...
         check_alpha_and_n_samples(alpha_np, n)
 
-        y_pred_proba = self.estimator_.predict(X, agg_scores)
-        y_pred_proba = self._check_proba_normalized(y_pred_proba, axis=1)
-        if agg_scores != "crossval":
-            y_pred_proba = np.repeat(
-                y_pred_proba[:, :, np.newaxis], len(alpha_np), axis=2
-            )
+        # Estimate prediction sets
+        prediction_sets = self.conformity_score_function_.predict_set(
```
candicemyt, line 588: predict_set function is not required, why not just call get_sets here ?

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,277 @@
...
+
+        Returns:
+        --------
+        NDArray
+            Array of quantiles with respect to alpha_np.
+        """
+        quantiles_ = 1 - alpha_np
+        return quantiles_
+
+    def _add_regualization(
```
candicemyt, line 124: typo :)

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,421 @@
...
+    check_include_last_label, check_proba_normalized
+)
+from mapie.estimator.classifier import EnsembleClassifier
+
+from mapie._machine_precision import EPSILON
+from mapie._typing import ArrayLike, NDArray
+from mapie.utils import compute_quantiles
+
+
+class APS(Naive):
```
candicemyt, line 18: Maybe APSConformityScore is more aligned with regression scores

mapie/classification.py
```diff
@@ -47,7 +42,7 @@ class MapieClassifier(BaseEstimator, ClassifierMixin):
         Method to choose for prediction interval estimates.
```
candicemyt, line 42: It would be useful for users to have comment like "warning depreciated use instead conformity_score_function_" comment here

thibaultcordier: Add keyword [DEPRECIATED] Done 👍  (line 71)

mapie/classification.py
```diff
@@ -979,91 +370,54 @@ def _check_fit_parameter(
...
-            - NDArray of shape (n_samples,)
-            - NDArray of shape (n_samples,)
-            - NDArray of shape (n_samples,)
-        """
-        # Split data for raps method
-        raps_split = StratifiedShuffleSplit(
-            n_splits=1, test_size=size_raps, random_state=self.random_state
+        cs_estimator = check_classification_conformity_score(
+            conformity_score=self.conformity_score,
+            method=self.method,
```
candicemyt, line 379: unecessary comma :)

mapie/conformity_scores/interface.py
```diff
@@ -0,0 +1,192 @@
...
+    def set_ref_predictor(
+        self,
+        predictor: BaseEstimator
+    ):
+        """
+        Set the reference predictor.
+
+        Parameters
+        ----------
+        predictor: BaeEstimator
```
candicemyt, line 42: typo :)

mapie/conformity_scores/interface.py
```diff
@@ -0,0 +1,192 @@
...
+
+        alpha_np: NDArray of shape (n_alpha, )
+            Represents the uncertainty of the confidence interval to produce.
+
+        **kwargs: dict
+            Additional keyword arguments.
+
+        Returns:
+        --------
+        The output strcture depend on the subclass.
```
candicemyt, line 190: typo on 'structure' :)

mapie/conformity_scores/regression.py
```diff
@@ -463,10 +372,38 @@ def get_bounds(
...
+
+        alpha_np: NDArray of shape (n_alpha, )
+            Represents the uncertainty of the confidence interval to produce.
+
+        **kwargs: dict
+            Additional keyword arguments.
+
+        Returns:
+        --------
+        The output strcture depend on the ``get_bounds`` method.
```
candicemyt, line 406: typo on "structure"

mapie/conformity_scores/classification.py
```diff
@@ -0,0 +1,224 @@
...
+        the estimator of type ``EnsembleClassifier`` and the conformity scores.
+
+        Parameters
+        ----------
+        X: NDArray of shape (n_samples, n_features)
+            Observed feature values.
+
+        alpha_np: NDArray of shape (n_alpha,)
+            NDArray of floats between 0 and 1, representing the uncertainty
+            of the confidence interval.
```
candicemyt, line 168: What do you mean by "confidence interval" here ?

mapie/conformity_scores/classification.py
```diff
@@ -0,0 +1,224 @@
...
+
+        alpha_np: NDArray of shape (n_alpha, )
+            Represents the uncertainty of the confidence interval to produce.
+
+        **kwargs: dict
+            Additional keyword arguments.
+
+        Returns:
+        --------
+        The output strcture depend on the ``get_sets`` method.
```
candicemyt, line 221: typo on "structure"

mapie/conformity_scores/sets/naive.py
```diff
@@ -0,0 +1,277 @@
...
+    def _get_last_included_proba(
+        self,
+        y_pred_proba: NDArray,
+        thresholds: NDArray,
+        include_last_label: Union[bool, str, None],
+        **kwargs
+    ) -> Tuple[NDArray, NDArray, NDArray]:
+        """
+        Function that returns the smallest score
+        among those which are included in the prediciton set.
```
candicemyt, line 159: typo on "prediction"

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,421 @@
...
+        predicition_sets: NDArray of shape (n_samples, n_alpha)
+            Prediction sets.
+
+        Returns:
+        --------
+        NDArray of shape (n_samples, n_alpha)
+            Vs parameters.
+        """
+        # compute V parameter from Romano+(2020)
+        vs = (
```
candicemyt, line 245: maybe v is a name more aligned with the function's signature

mapie/conformity_scores/sets/lac.py
```diff
@@ -0,0 +1,229 @@
...
+from mapie._typing import NDArray
+from mapie.utils import compute_quantiles
+
+
+class LACConformityScore(BaseClassificationScore):
+    """
+    Least Ambiguous set-valued Classifier (LAC) method-based
+    non conformity score (also formerly called ``"score"``).
+
+    It is based on the the scores (i.e. 1 minus the softmax score of the true
```
candicemyt, line 19: typo : "on the scores"

mapie/conformity_scores/sets/raps.py
```diff
@@ -0,0 +1,581 @@
...
+        groups: Optional[ArrayLike] of shape (n_samples,)
+            Group labels for the samples used while splitting the dataset into
+            train/test set.
+            By default ``None``.
+
+        Returns
+        -------
+        Tuple[NDArray, NDArray, NDArray, NDArray, Optional[NDArray],
+        Optional[NDArray]]
+            - NDArray of shape (n_samples, n_features)
```
candicemyt, line 128: Could you specify here what is each NDArray?

mapie/conformity_scores/classification.py
thibaultcordier: It should be replaced by "confidence set" 👍

mapie/conformity_scores/sets/aps.py
```diff
@@ -0,0 +1,421 @@
...
+            y_proba_last_cumsumed,
+            threshold,
+            y_pred_proba_last,
+            prediction_sets
+        )
+
+        # get random numbers for each observation and alpha value
+        random_state = check_random_state(self.random_state)
+        random_state = cast(np.random.RandomState, random_state)
+        us = random_state.uniform(size=(prediction_sets.shape[0], 1))
```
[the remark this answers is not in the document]
thibaultcordier: It should be replaced by "u_param" (it is just an intermediate uniform random variable).

mapie/regression/regression.py
thibaultcordier: After careful consideration, we're going to go with the first proposal.

candicemyt: -> we agree that this should not be done in this PR

candicemyt (approved):

mapie/regression/time_series_regression.py
```diff
@@ -114,7 +114,9 @@ def _relative_conformity_scores(
         """
         y_pred = super().predict(X, ensemble=ensemble)
         scores = np.array(
-            self.conformity_score_function_.get_conformity_scores(X, y, y_pred)
+            self.conformity_score_function_.get_conformity_scores(
+                y, y_pred, X=X
```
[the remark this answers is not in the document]
thibaultcordier: Because X is no longer the first parameter of the `get_conformity_score` method, but inside `**kwargs`, justifying its displacement.

PR #507, opened 2024-08-08
506 ccp reoptimize inference

### LINKED ISSUE #506: Conditional CP should re-optimize g at inference time
The CCP method related to #449 would require to correct the optimised values for each new X_n+1, considering a bound M, to have the expected theoretical guarantees  .

We need to re-optimize the g function, with an augmented dataset composed by the n calibration points and X_n+1. It can still be interesting to compute the optimisation with only the n calibration points, to use this optimised value as starting value for the inference optimizations.

Damien-Bouet:
# Description

I added a re-optimization for each X_n+1 at inference time (in the ``predict``). The old ``split`` method is still available with ``mapie.predict(X, unsafe_approximation=True)``.
Fixes #506 

## Type of change

- This change requires a documentation update (it was already updated in the doc)

# How Has This Been Tested?

- [x] Unit test
- [x] demo/tutorial notebook

# Checklist


examples/regression/4-tutorials/plot_ccp_tutorial.py
```diff
@@ -51,6 +53,14 @@
...
 Recall that the ``alpha`` is ``1 - target coverage``.
 
+Warning:
+
+In this tutorial, we use ``unsafe_approximation=True`` to have a faster
+computation (because Read The Docs examples require fast computation).
+This mode use an approximation, which make the inference (``predict``) faster,
+but induce a small miscoverage. It is recommanded not to use it, or be
+very careful and empirically check the coverage and a test set.
+
```
Damien-Bouet, line 63: I needed to temporally use the old method (with ``unsafe_approximation=True``), which doesn't have the theoretical guarantees, otherwise the documentation was in timeout. I will need to change the dataset, to have faster computation, or transform this .py tuto in the doc, into a true notebook, which is not run during the documentation build.
Indeed, showing in the first tuto, the deprecated method which doesn't have guarantees is not a good idea.

PR #574, opened 2024-12-18
CHORE: limit maximum sklearn version to avoid tests breaking, update maintainers list

Valentin-Laurent:
# Description

 - Limit sklearn version to avoid tests breaking with v1.6.0. Dependency support will be handled more properly with MAPIE v1.0.0
- Update maintainers list

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,15 +19,21 @@ PROJECT_URLS = {
     "Source Code": "https://github.com/scikit-learn-contrib/MAPIE"
 }
 LICENSE = "new BSD"
-MAINTAINER = "T. Cordier, V. Blot, L. Lacombe"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,15 +19,21 @@
...
+    "valentin.laurent@capgemini.com, "
     "thibault.a.cordier@capgemini.com, "
     "vincent.blot@capgemini.com, "
     "louis.lacombe@capgemini.com"
 )
 PYTHON_REQUIRES = ">=3.7"
 PACKAGES = find_packages()
-INSTALL_REQUIRES = ["scikit-learn", "scipy", "numpy>=1.21", "packaging"]
+INSTALL_REQUIRES = [
+    "scikit-learn < 1.6.0",
```
Valentin-Laurent, line 32:
```suggestion
    "scikit-learn<1.6.0",
```

PR #577, opened 2024-12-19
ENH: V1 CQR docstrings

jawadhussein462:


mapie_v1/regression.py
```diff
@@ -834,43 +834,43 @@ def predict(
...
-
-    confidence_level : Union[float, List[float]], default=0.9
+    estimator : Union[RegressorMixin, Pipeline,
+                    List[Union[RegressorMixin, Pipeline]]]
+        The base quantile regression model(s) used to estimate the target
+        quantiles.
+        - In `fit` mode (`prefit=False`), this should be a single quantile
+          regression estimator (e.g., `QuantileRegressor`) or a pipeline
+          combining preprocessing and regression.
+        - In `prefit` mode (`prefit=True`), this must be a list of three
```
Valentin-Laurent, line 855: Let's clarify the predit=True explanations here:

1. What type of models are supported
2. What are the lower, upper and median quantiles (ie must be (1-confidence_level)/2, confidence_level/2, 0.5)

PR #618, opened 2025-02-28
ENH: use pyproject.toml

### LINKED ISSUE #512: Switch from setup.py to pyproject.toml
**Context**
MAPIE is using setuptools (with `setup.py`) to build its distribution. This will soon have consequences, as pip 25.0 will enforce more modern tooling.

We should switch to pyproject.toml. [This ressource](https://github.com/MaartenGr/BERTopic/issues/1969) may be interesting. We should also use this opportunity to upgrade our `twine` and `wheel` dependencies to the latest.

**Describe the bug**
For example, installing MAPIE 0.9 in dev mode throws a pip deprecation warning. 

**To Reproduce**
Clone the repo then run `pip install -e .`

**Stacktrace**
```
DEPRECATION: Legacy editable install of MAPIE==0.9.0 from file:///Users/vlaurent/code/pro/MAPIE/MAPIE (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
  Running setup.py develop for MAPIE
Successfully installed MAPIE-0.9.0
```

**Additional info**
I got this email when uploading MAPIE to TestPyPi:

> In the future, PyPI will require all newly uploaded source distribution filenames to comply with [PEP 625](https://peps.python.org/pep-0625/). Any source distributions already uploaded will remain in place as-is and do not need to be updated.
> 
> Specifically, your recent upload of 'MAPIE-1.0.0a1.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'mapie'.
> 
> In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames.

C-BdB:
# Description

ENH: use pyproject.toml

Fixes #512 

## Type of change

# How Has This Been Tested?

- unit tests and coverage

# Checklist


doc/images/quickstart_1.png
Valentin-Laurent, line 1: Let's remove this change

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+    {name = "Thibault Cordier", email = "thibault.a.cordier@capgemini.com"},
+    {name = "Louis Lacombe", email = "louis.lacombe@capgemini.com"},
+    {name = "Vincent Blot", email = "vincent.blot@capgemini.com"},
+]
+requires-python = ">=3.7"
+dependencies = [
+    "scikit-learn<1.6.0",
+    "scipy",
+    "numpy>=1.21",
+    "packaging"
```
Valentin-Laurent, line 24: `packaging` was also present in the setup.py, but I'm wondering why we have this dependency. Any clue?

requirements.dev.txt
Valentin-Laurent, line 1: Let's update `environement.dev.yml` as well.

setup.py
```diff
@@ -1,71 +1,5 @@
...
-    download_url=DOWNLOAD_URL,
-    project_urls=PROJECT_URLS,
-    license=LICENSE,
-    maintainer=MAINTAINER,
-    maintainer_email=MAINTAINER_EMAIL,
-    packages=PACKAGES,
-    python_requires=PYTHON_REQUIRES,
-    install_requires=INSTALL_REQUIRES,
-    classifiers=CLASSIFIERS,
     zip_safe=False  # the package can run out of an .egg file
```
Valentin-Laurent, line 4: This flag seems obsolete (see [here](https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html)), shall we get rid of `setup.py` entirely? 😀

Valentin-Laurent (commented):
Thanks for the great work!
Few comments inline. Also, do we still need the MANIFEST.in?

pyproject.toml
C-BdB: Used for [compatibility tests](https://github.com/scikit-learn-contrib/MAPIE/blob/master/mapie/_compatibility.py).
We may delete it when deleting the tests on numpy<1.22.

setup.py
C-BdB: I prefer to let it as a very simple file, in case [we need it](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#what-if-something-that-can-not-be-changed-expects-a-setup-py-file)

C-BdB:
>  Also, do we still need the MANIFEST.in?

It's not a need but a I see that [some libraries](https://github.com/pandas-dev/pandas/blob/main/MANIFEST.in) still have it. [Other](https://github.com/scikit-learn/scikit-learn) don't. If we keep it, I suggest to update it with a mention to the notebooks:

```
include LICENSE
include AUTHORS.rst
recursive-exclude doc *
recursive-include examples *.py
recursive-include notebooks
```

setup.py
Valentin-Laurent: As discussed, let's remove it, and update the .bumpversion.cfg

Valentin-Laurent:
About `MANIFEST.in`: let's see what's included without. I think we shouldn't include /doc, /examples, or /notebooks.

C-BdB:
> About `MANIFEST.in`: let's see what's included without. I think we shouldn't include /doc, /examples, or /notebooks.

Here is the content of the dist/mapie-version.tar.gz built without setup.py nor MANIFEST.in:

LICENSE
PKG-INFO
mapie/
setup.cfg
AUTHORS.rst
MAPIE.egg-info/
README.rst
pyproject.toml

It seems ok.

Valentin-Laurent (changes requested):
Looks good 👍
Let's update HISTORY.rst

gmartinonQM:
Good Job !

HISTORY.rst
```diff
@@ -5,6 +5,10 @@ History
 0.9.x (2025-xx-xx)
 ------------------
 
+* Fix issue  512 to replace setup.py by pyproject.toml
```
Valentin-Laurent, line 8:
```suggestion
* Fix issue  512 to replace setup.py by pyproject.toml, bump twine and wheel dependencies to latest
```

Valentin-Laurent: (to keep the HISTORY.rst as concise as possible)

Valentin-Laurent (changes requested):

Valentin-Laurent (approved):

PR #625, opened 2025-03-18
DOC: Update `The conformity (or "calibration") set` page, with an emphasis on MAPIE methods rather than the internal works

Valentin-Laurent:


doc/images/cp_prefit.png
C-BdB, line 1: With the "Test data" out of the "MAPIE predict set" box, do you think the arrow from "predict_points" to "predict_sets (or intervals)" is still relevant?

Valentin-Laurent: Good point

Valentin-Laurent: @C-BdB I let you check the new images once the doc is built

C-BdB: I think images are clear.
Just one question: did you let the grey arrow on doc/images/cp_prefit.png on purpose?

Valentin-Laurent: Nope, fixed thx!

C-BdB (approved):

PR #665, opened 2025-04-23
V1 time series v1 wording

FaustinPulveric:


HISTORY.rst
```diff
@@ -71,7 +71,7 @@ History
 ------------------
 
 * Fix the quantile formula to ensure valid coverage for any number of calibration data in `ConformityScore`.
-* Fix overloading of the value of the `method` attribute when using `MapieRegressor` and `MapieTimeSeriesRegressor`.
```
Valentin-Laurent, line 74: We shouldn't change the history

mapie/regression/time_series_regression.py
```diff
@@ -299,17 +299,20 @@ def adapt_conformal_inference(
         X, y = cast(NDArray, X), cast(NDArray, y)
 
         self._get_alpha()
-        alpha = cast(Optional[NDArray], check_alpha(alpha))
```
Valentin-Laurent, line 302: Shouldn't we use `check_alpha` on `confidence_level`?
Not so much for the transformation, but at least for the checks.

mapie/regression/time_series_regression.py
```diff
@@ -299,17 +299,20 @@ def adapt_conformal_inference(
...
-        if alpha is None:
-            alpha = np.array(list(self.current_alpha.keys()))
-        alpha_np = cast(NDArray, alpha)
+        if confidence_level is None:
+            alpha_np = np.array(list(self.current_alpha.keys()))
+        else:
+            alpha_np = cast(
```
Valentin-Laurent, line 305: Is the casting mandatory here?

mapie/regression/time_series_regression.py
Valentin-Laurent: Also, maybe we should use confidence_level_np instead of alpha_np, because we end up using confidence level with .predict

mapie/regression/time_series_regression.py
```diff
@@ -392,19 +394,25 @@ def update(
...
+                X, y, ensemble=ensemble, confidence_level=confidence_level,
                 gamma=gamma, optimize_beta=optimize_beta
             )
         else:
             raise ValueError(
                 f"Invalid method. Allowed values are {self.valid_methods_}."
             )
 
-    def predict(
+    # For this ``predict`` function, we have chosen to ignore errors related to
```
Valentin-Laurent, line 405: Suggestion:

`# Overriding _MapieRegressor .predict method here. Bad practise, but this inheritance is questionable and will probably be reconsidered anyway.
`

mapie/regression/time_series_regression.py
```diff
@@ -454,16 +461,20 @@ def predict(
               - [:, 0, :]: Lower bound of the prediction interval.
               - [:, 1, :]: Upper bound of the prediction interval.
         """
-        if alpha is None:
```
Valentin-Laurent, line 457: Unsure about this refactoring of the .predict method, starting with the added `return`.

Valentin-Laurent (changes requested):

mapie/metrics/regression.py
```diff
@@ -476,15 +476,15 @@ def coverage_width_based(
     >>> y_preds_up = np.array([6, 9, 10, 12.5, 12])
     >>> eta = 0.01
     >>> alpha = 0.1
```
Valentin-Laurent, line 478: We can use confidence_level instead of alpha in the example (ie `confidence_level = 0.9`[...])

Valentin-Laurent (changes requested):

doc/v1_migration_guide.rst
```diff
@@ -243,8 +243,39 @@ Additionally, a number of classification and regression functions have been upda
...
+
+The ``MapieTimeSeriesRegressor`` class is now called ``TimeSeriesRegressor``.
+
+Functions update
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``adapt_conformal_inference``, ``update``, ``predict`` and ``coverage_width_based`` functions of the class now
+take ``confidence_level`` as input, instead of ``alpha`` (``confidence_level`` is equivalent to ``1 - alpha``).
+
+New function
```
Valentin-Laurent, line 271: That method is private (not meant to be used by users, so let's not mention it)

mapie/regression/time_series_regression.py
```diff
@@ -454,16 +454,29 @@ def predict(
...
 
         return super().predict(
             X, ensemble=ensemble, alpha=alpha, optimize_beta=optimize_beta,
             allow_infinite_bounds=allow_infinite_bounds, **predict_params
         )
+
+    # The public API changed from alpha to confidence_level.
+    # TODO: refactor this class to use confidence_level everywhere
+    @staticmethod
+    def transform_confidence_level_to_alpha_array(
```
Valentin-Laurent, line 474: Let's rename it to indicate it is private: `_transform_confidence_level_to_alpha_array`

Valentin-Laurent (changes requested):

PR #696, opened 2025-05-14
FIX: In the regression setting, cv=LeaveOneGroupOut() and cv=LeavePGroupsOut() are not working 

### LINKED ISSUE #542: In the regression setting, cv=LeaveOneGroupOut() and cv=LeavePGroupsOut() are not working
**Describe the bug**
In the regression setting, cv=LeaveOneGroupOut() and cv=LeavePGroupsOut() are not working

**To Reproduce**
Script to reproduce the behavior:

```python
from mapie.regression import MapieRegressor
from sklearn.datasets import load_iris
from sklearn.model_selection import *

X, y = load_iris(return_X_y=True)
groups = [0]*30 + [1]*30 + [2]*30 + [3]*30 + [4]*30

working = MapieRegressor(cv=GroupKFold())
working.fit(X, y, groups=groups)

not_working = MapieRegressor(cv=LeaveOneGroupOut()) # Same issue with LeavePGroupsOut(n_groups=n)
not_working.fit(X, y, groups=groups)
```

**Additional context**
The bug seems to come from this line of code, where `check_no_agg_cv` is not called with `groups`
`self.use_split_method_ = check_no_agg_cv(X, self.cv, self.no_agg_cv_)`

Simply adding groups here may solve the issue. Adding a test to prevent this error in the future would be nice.

sulphatet:
# Description

This pull request fixes an issue where the `_MapieRegressor` internal class did not support `LeaveOneGroupOut` and `LeavePGroupsOut` as a valid cross-validation strategy. This was due to incorrect passing of the `groups` parameter to the `_check_no_agg_cv` logic, which has now been corrected.

Fixes #542

## Type of change

# How Has This Been Tested?

- Added a new unit test: `test_group_cv_output_shape`, which:
    - [x] Imports and applies `LeaveOneGroupOut` and `LeavePGroupsOut` with groups passed to `.fit()`

    - [x] Asserts that predicted intervals have the correct shape

Note: Now that the `_MapieRegressor` is an internal class, classes like `SplitConformalRegressor` class in the v1 API wraps `_MapieRegressor` using `cv="prefit"`. This change does not directly affect it, but in future refactors, I think the handling of CV and group parameters could be carefully unified, given that `SplitConformalRegressor` inherits from `_MapieRegressor`.

# Checklist

- [ ] I have updated the [HISTORY.rst](https://github.com/scikit-learn-contrib/MAPIE/blob/master/HISTORY.rst) and [AUTHORS.rst](https://github.com/scikit-learn-contrib/MAPIE/blob/master/AUTHORS.rst) files
- [ ] When updating documentation: doc builds successfully and without warnings: `make doc`
- [ ] When updating documentation: code examples in doc run successfully: `make doctest`

### DIFF
--- a/mapie/estimator/regressor.py
+++ b/mapie/estimator/regressor.py
@@ -528,7 +528,12 @@ class EnsembleRegressor:
         **fit_params
     ) -> EnsembleRegressor:
 
-        self.use_split_method_ = _check_no_agg_cv(X, self.cv, self.no_agg_cv_)
+        self.use_split_method_ = _check_no_agg_cv(
+            X,
+            self.cv,
+            self.no_agg_cv_,
+            groups=groups
+            )
         single_estimator_: RegressorMixin
 
         if self.cv == "prefit":
--- a/mapie/tests/test_regression.py
+++ b/mapie/tests/test_regression.py
@@ -291,39 +291,27 @@ def test_predict_output_shape(
 
 
 @pytest.mark.parametrize(
-    "cv,   n_groups, n_leave",
+    "cv, n_groups",
     [
-        (LeaveOneGroupOut(),      5, 1),
-        (LeavePGroupsOut(2),     10, 2),
+        (LeaveOneGroupOut(), 5),
+        (LeavePGroupsOut(2), 10),
     ],
 )
-@pytest.mark.parametrize("alpha", [0.1, [0.05, 0.1]])
-def test_group_cv_output_shape(
-    cv: Any, n_groups: int, n_leave: int, alpha: Any
-) -> None:
+def test_group_cv_fit_runs_regressor(cv, n_groups):
     """
-    Regression: MapieRegressor must accept group‑based CV
-    splitters (LeaveOneGroupOut, LeavePGroupsOut) without
-    raising and must return outputs with the canonical shapes
-    (n_samples, 2, n_alpha).
+    `_MapieRegressor` should accept group‑based CV splitters
+    (LeaveOneGroupOut, LeavePGroupsOut) without raising.
     """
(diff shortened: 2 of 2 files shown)


Valentin-Laurent:
Hello @sulphatet, thank you for this contribution. I agree with your note, by the way.

Suggestion about the test: we could make it simply about not breaking when using LeaveOneGroupOut and LeavePGroupsOut, without the parametrization of alpha and verifications on output shape (so no predict needed).

We have a lot of "end-to-end" tests like that in MAPIE, but they end up making the test suite quite long to run.

What do you think?

sulphatet:
Thanks for the feedback @Valentin-Laurent that makes sense, and I’ve simplified the test as suggested to only check that `.fit()` works without running `.predict()`. Let me know if this change is more in line what you were thinking. I'd be happy to make any further modifications.

Additionally, since `_MapieRegressor` is now internal and being inherited by classes like `SplitConformalRegressor`, I think we’ll eventually also want to ensure group-based CV is _**fully**_ tested through those public APIs?

Valentin-Laurent:
Looks good to be, thank you 👍

> we’ll eventually also want to ensure group-based CV is fully tested through those public APIs
That's right.

However, having too many functional ("end-to-end") tests is not ideal (it makes the test suite huge and slow if we want to cover all possible scenarios). Actually we already have functional tests in `tests_v1/test_functional/test_non-regression...` covering groups, so I'd say we're fine for now.

In the future, I'd rather focus on having a more modular codebase, so we can better unit test and catch issues like the one you just fixed.

Valentin-Laurent (approved):

PR #722, opened 2025-07-15
ENH: update HISTORY.rst

FaustinPulveric:


HISTORY.rst
```diff
@@ -5,6 +5,8 @@ History
 1.0.1 (2025-05-22)
 ------------------
 
+* Fix double inference when using `predict_set` function in split conformal classification
```
Valentin-Laurent, line 8: Should be above the current version (you can have a look at past versions of this file for an example)

PR #729, opened 2025-07-23
MTN: update HISTORY.rst

FaustinPulveric:


HISTORY.rst
```diff
@@ -7,6 +7,7 @@ History
 
 * Fix double inference when using `predict_set` function in split conformal classification
 * Add FAQ entry in the documentation about ongoing works to extend MAPIE for LLM control
+* Lift restrictions on Python versions 3.12 and 3.13
```
Valentin-Laurent, line 10: Thanks for the PR. Let's make it crystal clear for users: something like "MAPIE now supports Python versions up to the latest (currently 3.13)", or something like that

Valentin-Laurent (approved):

PR #734, opened 2025-07-29
FIX: prefit=True by default in split methods docstrings

FaustinPulveric:
There was an inconsistency in the default value of the ``prefit`` parameter in split methods (regression and classification) between the implementations and the docstrings.
The ``prefit`` parameter is now set to ``True`` by default in both.


Valentin-Laurent (changes requested):
LGTM, let's just add en entry in HISTORY.rst

PR #745, opened 2025-08-21
Fix warnings when running tests

FaustinPulveric:


mapie/aggregation_functions.py
```diff
@@ -112,8 +112,12 @@ def aggregate_all(agg_function: Optional[str], X: NDArray) -> NDArray:
     array([14.5, 14.5])
 
     """
-    if agg_function == "median":
-        return np.nanmedian(X, axis=1)
-    elif agg_function == "mean":
-        return np.nanmean(X, axis=1)
+    row_nan_mask = np.isnan(X).all(axis=1)
```
Valentin-Laurent, line 115: If I understand correctly, the warning happens when the entire axis is NaN is that right?

mapie/utils.py
```diff
@@ -635,7 +635,7 @@ def _check_alpha_and_n_samples(
     else:
         alphas_ = alphas
     for alpha in alphas_:
-        if n < np.max([1/alpha, 1/(1-alpha)]):
+        if alpha == 0 or alpha == 1 or n < np.max([1/alpha, 1/(1-alpha)]):
```
Valentin-Laurent, line 638: To me that change is not trivial:

- Maybe checking for alpha == 1 or 0 is done (or should be done) earlier in the code
- If we decide to check this, we need to update the docstring and the error message as well

PR #761, opened 2025-09-24
MTN: improve risk control tests

allglc:


Valentin-Laurent (commented):
Looks pretty good! Review for the commit "add dummy constants". Haven't checked the following ones yet.
To push the constant thing further, we could have created a `dummy_risk`, but that's probably a bit overkill.

Valentin-Laurent (approved):

PR #763, opened 2025-10-01
DOC : Correct some docstrings and Improve Quickstart for risk control binary classification

hmaissoro:


examples/risk_control/1-quickstart/plot_risk_control_binary_classification.py
```diff
@@ -21,17 +21,71 @@
...
+)
+
+plt.suptitle("Visualization of Train, Calibration, and Test Sets", fontsize=22)
+plt.tight_layout(rect=[0, 0.05, 1, 0.95])
+plt.show()
+
+##############################################################################
+# Second, fit a KNeighborsClassifier on the training data.
+
+# Fit KNeighborsClassifier on training data
```
Valentin-Laurent, line 87: Duplicate comment in regard to the markdown cell above

Valentin-Laurent (changes requested):
Good work!
A comment here and another below: if you take a look at the diff, your VSCode formatter has changed some irrelevant lines. Ideally, we commit only relevant changes, and not formatting. But let's move on and forget about it for now :)

examples/risk_control/1-quickstart/plot_risk_control_binary_classification.py
hmaissoro: Got it, thank your for your comments

PR #771, opened 2025-10-07
UPD: update Calibrate return

hmaissoro:


mapie/risk_control.py
```diff
@@ -1020,7 +1020,8 @@ def calibrate(  # pragma: no cover
 
         Returns
         -------
-        None
+        BinaryClassificationController
+            The fitted controller instance (for chaining).
```
Valentin-Laurent, line 1024: *calibrated

PR #784, opened 2025-10-30
Passed Predict Params to RAPS for conformity score calculation and EnsembleClassifier

### LINKED ISSUE #614: predict_params is not passed everywhere in MapieClassifier in some settings
Predict params are not passed to the underlying classifiers in some settings:

- [ ] In `mapie/estimator/classifier.py`, in the prefit setting, we should pass predict_params. Here is the code to fix:

```python
if self.cv == "prefit":
            y_pred_proba = self.single_estimator_.predict_proba(X)
            y_pred_proba = self._check_proba_normalized(y_pred_proba)
```

- [ ] In `mapie/conformity_scores/sets/raps.py`, same issue. Fixing this is a bit trickier since the `get_conformity_scores` score method currently doesn't have access to `predict_params`. Here is the code to fix:

```python
        self.y_pred_proba_raps = (
            self.predictor.single_estimator_.predict_proba(self.X_raps)
        )
```

- [x] We should clarify in the doctstring that `predict_params` are passed to both `predict` and `predict_proba` methods. This may be subject to discussion as this implementation choice may not be relevant.
- [ ] Optional: implement tests to check that the above works

VishnoiAman777:
# Description

This change adds a test case to ensure that `predict_params` are correctly passed and handled when using `Ensemble Classifier` to `predict_proba` function for an estimator and  `RAPSConformityScore` Class. Addtional test `test_raps_with_predict_params` is added to verify that the `predict_params` are properly passed through the `fit` and `predict` methods in RAPSConformityScore.

Fixes #614 

## Type of change

# How Has This Been Tested?

The following test was added to verify the changes:

- [x] `test_raps_with_predict_params`
   - This test verifies that `predict_params` are correctly passed through the `fit` and `predict` methods when using `RAPSConformityScore`

# Checklist

### DIFF
--- a/mapie/conformity_scores/sets/raps.py
+++ b/mapie/conformity_scores/sets/raps.py
@@ -180,15 +180,10 @@ class RAPSConformityScore(APSConformityScore):
             Conformity scores.
         """
         # Compute y_pred and position on the RAPS validation dataset
-        predict_params = kwargs.pop("predict_params", None)
-        if predict_params is not None and len(predict_params) > 0:
-            self.y_pred_proba_raps = self.predictor.single_estimator_.predict_proba(
-                self.X_raps, **predict_params
-            )
-        else:
-            self.y_pred_proba_raps = self.predictor.single_estimator_.predict_proba(
-                self.X_raps
-            )
+        predict_params = kwargs.pop("predict_params", {})
+        self.y_pred_proba_raps = self.predictor.single_estimator_.predict_proba(
+            self.X_raps, **predict_params
+        )
         self.position_raps = get_true_label_position(
             self.y_pred_proba_raps, self.y_raps
         )
--- a/mapie/tests/test_classification.py
+++ b/mapie/tests/test_classification.py
@@ -1717,7 +1717,8 @@ def test_raise_error_new_class() -> None:
     """
     clf = LogisticRegression()
     clf.fit(X, y)
-    y[-1] = 10
+    y_modified = y.copy()
+    y_modified[-1] = 10
     mapie_clf = _MapieClassifier(
         estimator=clf,
         conformity_score=APSConformityScore(),
@@ -1725,7 +1726,7 @@ def test_raise_error_new_class() -> None:
         random_state=random_state,
     )
     with pytest.raises(ValueError, match=r".*Values in y do not matched values.*"):
-        mapie_clf.fit(X, y)
+        mapie_clf.fit(X, y_modified)
 
 
 def test_fit_parameters_passing() -> None:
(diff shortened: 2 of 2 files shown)


mapie/conformity_scores/sets/raps.py
```diff
@@ -180,9 +180,15 @@ def get_conformity_scores(
             Conformity scores.
         """
         # Compute y_pred and position on the RAPS validation dataset
-        self.y_pred_proba_raps = self.predictor.single_estimator_.predict_proba(
-            self.X_raps
-        )
+        predict_params = kwargs.pop("predict_params", None)
```
allglc: Don't you only need to add `**predict_params` to the function? i.e,:
`self.y_pred_proba_raps = self.predictor.single_estimator_.predict_proba(self.X_raps, **predict_params)` like you did in the `classifier.py` file.

mapie/tests/test_classification.py
```diff
@@ -1764,6 +1764,42 @@ def test_predict_parameters_passing() -> None:
     np.testing.assert_equal(y_pred, 0)
 
 
+def test_raps_with_predict_params() -> None:
```
allglc: This test does not seem to test what we want. It does not test if predict params are passed into RAPS and actually this test passes on the current version of MAPIE (i.e., without your modifications).

Furthermore, the code is too complex: there is no need to recreate a dataset. Please refer to e.g., `test_predict_parameters_passing` for inspiration.

mapie/conformity_scores/sets/raps.py
VishnoiAman777: Hi @allglc, conformity_score_function_.get_conformity_scores() method also accepts additional parameters beyond the basic inputs. Specifically, it includes:

1. X
2. sample_weight
3. groups
4. predict_params

To ensure these are properly handled, I explicitly extracted predict_params from kwargs in the implementation.
```@python
        self.conformity_scores_ = self.conformity_score_function_.get_conformity_scores(
            y,
            y_pred_proba,
            y_enc=y_enc,
            X=X,
            sample_weight=sample_weight,
            groups=groups,
            predict_params=predict_params,
        )
```

mapie/tests/test_classification.py
VishnoiAman777: Apologies for the oversight—I should have reviewed this more thoroughly earlier. I’ve added a few commits to make the implementation more robust.

Additionally, I noticed that the test `test_raise_error_new_class` modifies the global X and y variables from make_classification. Since this can cause issues if the test runs before others, I’ve created a local copy of y within the function to avoid unintended side effects.

VishnoiAman777:
Hi @allglc, @LacombeLouis, can you please provide some feedback if there are changes required in the above implementation.

PR #788, opened 2025-11-04
ENH: add ability to handle multi-dimensional thresholds

allglc:
# Description

Add the ability to handle multi-dimensional thresholds (lambdas)

- [x] 'predict_function` can now also be a general function (X *params) -> 0/1
- [x] `predict_params` is now an argument (even when only one-dimensional lambda) and docstring should be clearer
- [x] `best_predict_param` is a tuple for multi-dimensional parameters
- [x] added an automatic flag is_multi_dimensional_param (based on `predict_params` dimension)
- [x] `_get_predictions_per_param` handles general `predict_functions`, but will process all parameter values sequentially (i don't know how to do it easily in parallel).
- [x] `get_predictions_per_param` will check the prediction values in the calibration step (using a new argument `is_calibration_step`, because I don't want to check at test time as it is not necessary and it might happen that when predicting a single probability, which can happen at test time, a value of 0 or 1 and would raise a warning)
- [x] predictions are checked so that for one-dimensional parameters, predictions should not be 0 or 1 all the time, and for multi-dimensional parameters, they should be 0 or 1 all the time.
- [x] I don't think it's necessary to add a custom error message when the dimension of the parameters do not match the inputs of the general predict function, as the default message seems explicit (cf `test_error_multi_dim_params_dim_mismatch`)

## To manage the two types of predict functions in `__init__` there are a few options:
1. Manual flag in the arguments of `BinaryClassificationController`
2. auto detection based on the function signature (does it takes one or more arguments) but it can cause issues when the predict_proba can also take several arguments (e.g. XGBoost)
3. add another argument e.g., ` predict_function_general` and the user has to provide at least this or the original `predict_function`.
4. Automatic check of predict_params.shape[1] > 1: the user has to provide a custom array of predict_params to do multi params

I will go with option 4.

### DIFF
--- a/examples/risk_control/2-advanced-analysis/plot_risk_control_multi_parameter_binary_classification.py
+++ b/examples/risk_control/2-advanced-analysis/plot_risk_control_multi_parameter_binary_classification.py
@@ -3,11 +3,11 @@
 Use MAPIE to control risk of a binary classifier with multiple prediction parameters
 ====================================================================================
 
-AI is a powerful tool for mail sorting (for example between spam and urgent mails).
-However, because algorithms are not perfects,it sometimes requires manual verification.
-Thus one would like to be able to controlthe amount of mail sent to human validation.
-One way to do so is to define a multi-parameters prediction function based on a
-classifier predicted scores. This would allow to define a rule for mail checking,
+AI is a powerful tool for email sorting (for example between spam and urgent emails).
+However, because algorithms are not perfect, manual verification is sometimes required.
+Thus one would like to be able to control the amount of emails sent to human validation.
+One way to do so is to define a multi-parameter prediction function based on a
+classifier's predicted scores. This would allow defining a rule for email checking,
 which could be adapted by varying the prediction parameters.
 
 In this example, we explain how to do risk control for binary classification relying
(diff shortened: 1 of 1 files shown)


mapie/risk_control/binary_classification.py
```diff
@@ -350,31 +376,47 @@ def _get_risk_values_and_eff_sample_sizes(
...
+        if self.is_multi_dimensional_param:
+            y_pred = np.empty((n_params, n_samples))
+            for i in range(n_params):
+                y_pred[i] = self._predict_function(X, *params[i])
+            if is_calibration_step:
+                self._check_predictions(y_pred)
+            y_pred = y_pred.astype(int)
+        else:
+            try:
+                predictions_proba = self._predict_function(X)[:, 1]
```
GBrelurut: Y-a-t-il une raison pour ne pas mettre la fonction de prédiction en multiparam dans le try ?

allglc: Oui car c'est pas le même appel de fonction et la deuxième erreur n'est pas adaptée (on peut peut-être adapter pour quand même tester la première erreur en multi-dim)

PR #802, opened 2025-11-18
DEV-XP: handle expected RuntimeWarnings in unit tests

hmaissoro:


mapie/tests/test_classification.py
```diff
@@ -1407,6 +1407,7 @@ def test_image_cumulated_scores(X: Dict[str, ArrayLike]) -> None:
     np.testing.assert_allclose(y_ps[:, :, 0], cumclf.y_pred_sets)
 
 
+@pytest.mark.filterwarnings("ignore:: UserWarning")
 @pytest.mark.parametrize("y_pred_proba", Y_PRED_PROBA_WRONG)
```
GBrelurut: rational : we ant to test the raising of the error, not the warning, so we ignore it

PR #811, opened 2025-11-28
chore: Create a new command for long tests and exclude them from standard coverage testing.

hmaissoro:


allglc:
I'm testing pytest-xdist to run tests in parallel. However there was an error with mypy: it required to use python >= 3.10. When doing this (in mypy.ini), there are new typing errors. So for testing purposes, I removed this step from the CI, just to see how much faster pytest-xdist is. It does not seem to improve much on GitHub runners. I'll have to test again when tests pass (they currently fail) and self hosted runner is free.

PR #813, opened 2025-12-01
docs: enhance multi-parameter plot with valid zones and best parameter

hmaissoro:


allglc:
Did a rebase to avoid erasing the modifications about using the new "predicted_positive"_fraction"risk

PR #830, opened 2025-12-16
refactor: improve the readability of 'find_precision_best_predict_param' method

hmaissoro:


mapie/risk_control/methods.py
```diff
@@ -418,8 +418,10 @@ def find_precision_best_predict_param(
             l_best_predict_param.append(np.nan)
             l_r_star.append(np.nan)
         else:
-            idx = np.argmin(valid_index[i])
-            l_best_predict_param.append(lambdas[valid_index[i][idx]])
-            l_r_star.append(r_hat[valid_index[i][idx]])
+            idx_min: int = int(np.min(valid_index[i]))
```
allglc: why doing ": int" typing ?

PR #834, opened 2025-12-17
Semantic segmentation controller

vincentblot28:
New class which inherits from MultilabelClassificationController to contorl precision and recall in semantic segmentation tasks


mapie/conformity_scores/regression.py
```diff
@@ -208,7 +208,7 @@ def _beta_optimize(
             "https://github.com/scikit-learn-contrib/MAPIE/issues/588"
         )
 
-        beta_np = np.full(
+        beta_np: NDArray[np.int_] = np.full(
```
allglc: I would prefer using a cast as it would enter the list of typing issues we have to fix. The current way will be unnoticed when we will work on fixing typing issues, while using cast make it easy to find and fix them.

PR #855, opened 2026-01-20
feat: add abstention support to BinaryClassificationController

hmaissoro:
- Added abstention support to `BinaryClassificationController`
- Added new `BinaryClassificationRisk` instances for abstention: `precision_positive`, `precision_negative`, and `abstention_rate`
- Strengthened validation of `best_predict_param_choice`

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -15,6 +15,7 @@ History
 * Add a disclosure about LLM usage to the pull request template.
 * Fix data leakage for time series tutorials
 * Improve documentation display (increase width and improve API table)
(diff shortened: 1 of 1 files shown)


mapie/risk_control/risks.py
```diff
@@ -154,3 +154,21 @@ def get_value_and_effective_sample_size(
     risk_condition=lambda y_true, y_pred: np.repeat(True, len(y_true)),
     higher_is_better=False,
 )
+
+precision_positive = BinaryClassificationRisk(
```
allglc: same as precision. is it necessary to implement?

PR #880, opened 2026-03-13
chore: fix all cast

allglc:
# Description

### Summary

- **Removed all `typing.cast` usages** from the `mapie` package (code, tests, and examples) while keeping both the default and minimal-dependency type-checks passing. This was done by tightening return types, using `np.asarray`, and adding targeted `assert` / `isinstance`-based narrowing instead of casts.
- **Improved type helpers and interfaces**:
  - `_check_alpha` now returns `Optional[NDArray]`, and all callers use that concrete type (no casts on its result).
  - Added `_column_or_1d_ndarray` to wrap `sklearn.utils.validation.column_or_1d` and always return an `NDArray`, replacing all `cast(NDArray, column_or_1d(...))` call sites.
  - Rewrote `_cast_point_predictions_to_ndarray` and `_cast_predictions_to_ndarray_tuple` to rely on `isinstance` checks and `assert` for narrowing instead of casts.
- **Strengthened type narrowing and conversions at boundaries**:
  - Replaced many internal `cast(NDArray, X/y/...)` occurrences with a *single* boundary conversion via `np.asarray(...)` in classifiers, regressors, calibration, risk control, and conformity scores; optional arrays (e.g. `sample_weight`, `groups`) are converted only when not `None`.
  - For cross-validation objects, replaced `cast(BaseCrossValidator, cv)` with `assert isinstance(cv, (BaseCrossValidator, BaseShuffleSplit))` and used the narrowed `cv`.
  - In residual-based utilities and helpers, removed `cast(ArrayLike, X_)` and now pass/return `np.asarray(...)` where needed.
- **Bug fixes and edge-case handling**:
  - **RAPS regularization** (`mapie/conformity_scores/sets/raps.py`):
    - Fixed the bug in `_add_regularization` where `k_star` was incorrectly derived from `lambda_`.
    - Carefully handle `lambda_star` / `k_star` whether they are scalars or arrays, using intermediate variables and `np.asarray(...).item(0)` where needed; where old numpy stubs are too generic, added narrowly scoped `# type: ignore[assignment]` instead of reintroducing `cast`.
  - **Risk control**: made `effective_sample_size` and `_delta` handling explicitly integral/float using `int(...)` / `float(...)` plus asserts, rather than casts.
  - **Random state handling**: introduced `_random_seeds_from_state()` so that `np.random.get_state()` is accessed in one typed place, and replaced the previous cast-based indexing with this helper.
  - **Example script** (`examples/regression/2-advanced-analysis/plot_timeseries_enbpi.py`): removed `cast` and now use `np.asarray(result["y_pis"])`.
- **Compatibility with minimal requirements**:
  - Verified mypy passes under both the project’s dev environment and under the **minimal dependency constraints** (Python 3.9, `numpy==1.24.1`, `scikit-learn==1.4.*`, `scipy==1.10.*`) using:
    - `uv run --python 3.9 --with numpy==1.24.1 --with "scikit-learn==1.4.*" --with scipy==1.10.* --with mypy mypy mapie`
  - Where older numpy type stubs were overly generic (notably in RAPS), used very localized `# type: ignore[...]` comments on assignments to avoid reintroducing `cast()`.

Overall, this PR removes `cast()` calls across the codebase, replaces them with more precise typing and runtime checks, and keeps static typing green for both standard and minimal-dependency environments.

## LLM Usage

If you used an LLM, please provide the following details:

- **LLM used**: Cursor
- **Purpose**:code generation

### DIFF
--- a/examples/regression/2-advanced-analysis/plot_timeseries_enbpi.py
+++ b/examples/regression/2-advanced-analysis/plot_timeseries_enbpi.py
@@ -31,12 +31,10 @@ to the target, along with narrower PIs.
 """
 
 import warnings
-from typing import cast
 
 import numpy as np
 import pandas as pd
 from matplotlib import pylab as plt
-from numpy.typing import NDArray
 from scipy.stats import randint
 from sklearn.ensemble import RandomForestRegressor
 from sklearn.model_selection import PredefinedSplit, RandomizedSearchCV
@@ -210,7 +208,7 @@ for i, (ax, w, result) in enumerate(
         label="Predictions",
     )
 
-    y_pis = cast(NDArray, result["y_pis"])
+    y_pis = np.asarray(result["y_pis"])
 
     ax.fill_between(
         demand_test.index,
--- a/mapie/calibration.py
+++ b/mapie/calibration.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import warnings
 from inspect import signature
-from typing import Dict, Optional, Tuple, Union, cast
+from typing import Dict, Optional, Tuple, Union
 
 import numpy as np
 from numpy.typing import ArrayLike, NDArray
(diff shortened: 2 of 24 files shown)


GBrelurut:
The package as grown a lot since we initiated this branch. We handle the problem in continue.

PR #881, opened 2026-03-14
feat: expose `p0_p1_output` in `VennAbersCalibrator.predict_proba`

### LINKED ISSUE #870: `VennAbersCalibrator.predict_proba` does not expose `p0_p1_output` parameter
Hi team,

I noticed that `VennAbersCalibrator` relies internally on the `VennAbers` class. The underlying `VennAbers.predict_proba` method supports the p0_p1_output parameter, which allows returning the (p0, p1) pair instead of the averaged calibrated probability.

In the current MAPIE implementation, however, this parameter is not exposed in `VennAbersCalibrator.predict_proba`. It appears that `p0_p1_output` is effectively fixed to False internally, and users cannot control this behavior through the public API.

**Why this matters**
In some workflows, it can be useful to access the raw (p0, p1) outputs rather than only the averaged calibrated probability. For example, the interval (p0, p1) may be reported alongside the calibrated probability as a measure of uncertainty.

Since the underlying `VennAbers` implementation already supports this functionality, exposing the `p0_p1_output` parameter in the MAPIE wrapper would increase flexibility without altering default behavior.

Would it be possible to expose this parameter in `VennAbersCalibrator.predict_proba`?

Please let me know if I have misunderstood the intended design or missed something in the implementation.

Thanks for your work on MAPIE!

LEDazzio01:
## Description

Closes #870

The internal `VennAbersCV.predict_proba` and `VennAbersMultiClass.predict_proba` already support a `p0_p1_output` parameter to return raw (p0, p1) multiprobability outputs. However, `VennAbersCalibrator.predict_proba` was hardcoding `p0_p1_output=False` and never exposing this parameter to users.

This PR:
- Adds `p0_p1_output: bool = False` parameter to `VennAbersCalibrator.predict_proba`
- Passes it through to all three code paths: prefit binary, prefit multiclass, and IVAP/CVAP
- When `True`, returns `Tuple[NDArray, p0_p1_data]` instead of just `NDArray`
- Updates the docstring with the new parameter and return type documentation

## Changes

- **`mapie/calibration.py`**: Modified `VennAbersCalibrator.predict_proba` signature and all internal call sites
- Added `Any` to typing imports for the return type annotation

## Testing

- All 102 existing `test_calibration.py` tests pass ✅
- Manually verified `p0_p1_output=True` works correctly for:
  - Binary IVAP
  - Multi-class IVAP
  - Prefit binary
  - Prefit multi-class

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -52,5 +52,6 @@
 * Omid Gheibi <omidgheibi@gmail.com>
 * Aman Vishnoi <amanvishnoi777@gmail.com>
 * Hannes Körner <HannesMK>
+* Elaine Dias <LEDazzio01>
 
 To be continued ...
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -52,6 +52,6 @@
 * Omid Gheibi <omidgheibi@gmail.com>
 * Aman Vishnoi <amanvishnoi777@gmail.com>
 * Hannes Körner <HannesMK>
-* Elaine Dias <LEDazzio01>
+* L. Elaine Dazzio <LEDazzio01>
 
 To be continued ...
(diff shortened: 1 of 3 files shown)


allglc (approved):
All good for me!

PR #882, opened 2026-03-14
fix: correctly route `sample_weight` through sklearn Pipeline in `_fit_estimator`

### LINKED ISSUE #798: Unable to pass sample weights properly to MAPIE Underlying classifier and regressor
**Describe the bug**
Unable to use Sklearn pipeline estimators with `SplitConfomalRegressor` and `SplitConformalClassifier`. Sample weights  are not passed to appropriately pipeline estimator in case of fit in case of `CrossConformalRegressor` and `CrossConfomalClassifier`. 

**To Reproduce**
```@python
import numpy as np
from numpy.typing import NDArray
from sklearn.neural_network import MLPRegressor
from mapie.metrics.regression import regression_coverage_score
from mapie.regression import SplitConformalRegressor
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import PolynomialFeatures
from sklearn.linear_model import LinearRegression
from mapie.utils import train_conformalize_test_split

RANDOM_STATE = 42

X, y = make_regression(
    n_samples=1000, n_features=10, noise=20, random_state=RANDOM_STATE
)
(X_train, X_conformalize, X_test,
 y_train, y_conformalize, y_test) = train_conformalize_test_split(
    X, y,
    train_size=0.8, conformalize_size=0.1, test_size=0.1,
    random_state=RANDOM_STATE
)
sample_weight_train = np.random.rand(len(X_train))

pipeline = Pipeline([
    ('poly_features', PolynomialFeatures(degree=2)),
    ('linear_regression', LinearRegression())
])

confidence_level = 0.95
mapie_regressor = SplitConformalRegressor(
    estimator=pipeline, confidence_level=confidence_level, prefit=False
)
mapie_regressor.fit(X_train, y_train, {"sample_weight": sample_weight_train})
```

Error:
```@python
ValueError: Pipeline.fit does not accept the sample_weight parameter. You can pass parameters to specific steps of your pipeline using the stepname__parameter format, e.g. `Pipeline.fit(X, y, logisticregression__sample_weight=sample_weight)`.
```

Further in case of `CrossConformalRegressor` and `CrossConformalClassifer` sample_weights are completely ignored because `EnsembleRegressor` and `EnsembleClassifier` fit's estimator using this function present in utils.py
```@python
def _fit_estimator(
    estimator: Union[RegressorMixin, ClassifierMixin],
    X: ArrayLike,
    y: ArrayLike,
    sample_weight: Optional[NDArray] = None,
    **fit_params,
) -> Union[RegressorMixin, ClassifierMixin]:
    fit_parameters = signature(estimator.fit).parameters
    supports_sw = "sample_weight" in fit_parameters
    if supports_sw and sample_weight is not None:
        estimator.fit(X, y, sample_weight=sample_weight, **fit_params)
    else:
        estimator.fit(X, y, **fit_params)
    return estimator
```
Since pipeline object don't have sample_weight as `estimator.fit parameters`, so `support_sw` will always be None in case of pipeline estimator.

**Expected behavior**
I should be able to pass sample_weights in case of `prefit=False` case to `SplitConformalRegressor`, and `SplitConfromalClassifer`. For `CrossConformalRegressor` and `CrossConformalClassifer` I should pass sample_weight to each of the estimator that could use it. Also we should implement test cases for the above classes.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - MAPIE Version: v1.1.0

LEDazzio01:
## Description

Closes #798

When the estimator passed to `_fit_estimator` is a sklearn `Pipeline`, the function previously inspected `signature(estimator.fit).parameters` for `"sample_weight"`. Since `Pipeline.fit()` uses `**fit_params` and doesn't explicitly list `sample_weight` in its signature, `supports_sw` was always `False`, silently dropping sample weights.

## Root Cause

```python
# Before: checks Pipeline.fit() signature — no sample_weight there
fit_parameters = signature(estimator.fit).parameters
supports_sw = "sample_weight" in fit_parameters  # Always False for Pipeline
```

## Fix

When the estimator is a `Pipeline`, inspect the **final step's** `fit` signature instead, and route `sample_weight` using sklearn's `stepname__sample_weight` convention:

```python
if isinstance(estimator, Pipeline):
    final_step = estimator[-1]
    final_step_name = estimator.steps[-1][0]
    fit_parameters = signature(final_step.fit).parameters
    supports_sw = "sample_weight" in fit_parameters
    if supports_sw and sample_weight is not None:
        sw_key = f"{final_step_name}__sample_weight"
        estimator.fit(X, y, **{sw_key: sample_weight}, **fit_params)
```

## Changes

- **`mapie/utils.py`**: Modified `_fit_estimator` to handle `Pipeline` estimators (`Pipeline` was already imported)
- Updated docstring to document Pipeline behavior

## Testing

- All 142 existing `test_utils.py` tests pass ✅
- Verified Pipeline + sample_weight produces identical results to plain estimator + sample_weight:
  - `LinearRegression` weighted coef = 1.940120
  - `Pipeline(PolynomialFeatures, LinearRegression)` weighted coef = 1.940120 ✅

### DIFF
--- a/mapie/classification.py
+++ b/mapie/classification.py
@@ -34,6 +34,7 @@ from mapie.utils import (
     _check_null_weight,
     _check_predict_params,
     _check_verbose,
+    _fit_estimator,
     _prepare_fit_params_and_sample_weight,
     _prepare_params,
     _raise_error_if_fit_called_in_prefit_mode,
@@ -179,8 +180,8 @@ class SplitConformalClassifier:
         _raise_error_if_method_already_called("fit", self._is_fitted)
 
         cloned_estimator = clone(self._estimator)
-        fit_params_ = _prepare_params(fit_params)
-        cloned_estimator.fit(X_train, y_train, **fit_params_)
+        fit_params_, sample_weight = _prepare_fit_params_and_sample_weight(fit_params)
+        _fit_estimator(cloned_estimator, X_train, y_train, sample_weight, **fit_params_)
         self._mapie_classifier.estimator = cloned_estimator
 
         self._is_fitted = True
--- a/mapie/regression/regression.py
+++ b/mapie/regression/regression.py
@@ -32,6 +32,7 @@ from mapie.utils import (
     _check_null_weight,
     _check_predict_params,
     _check_verbose,
+    _fit_estimator,
     _prepare_fit_params_and_sample_weight,
     _prepare_params,
     _raise_error_if_fit_called_in_prefit_mode,
@@ -177,13 +178,8 @@ class SplitConformalRegressor:
         _raise_error_if_method_already_called("fit", self._is_fitted)
 
         cloned_estimator = clone(self._estimator)
-        fit_params_ = _prepare_params(fit_params)
-        if isinstance(cloned_estimator, Pipeline) and "sample_weight" in fit_params_:
-            final_step_name = cloned_estimator.steps[-1][0]
-            fit_params_[f"{final_step_name}__sample_weight"] = fit_params_.pop(
-                "sample_weight"
-            )
-        cloned_estimator.fit(X_train, y_train, **fit_params_)
+        fit_params_, sample_weight = _prepare_fit_params_and_sample_weight(fit_params)
+        _fit_estimator(cloned_estimator, X_train, y_train, sample_weight, **fit_params_)
         self._mapie_regressor.estimator = cloned_estimator
 
         self._is_fitted = True
--- a/mapie/regression/regression.py
+++ b/mapie/regression/regression.py
@@ -32,7 +32,6 @@ from mapie.utils import (
     _check_null_weight,
     _check_predict_params,
     _check_verbose,
-    _fit_estimator,
     _prepare_fit_params_and_sample_weight,
     _prepare_params,
     _raise_error_if_fit_called_in_prefit_mode,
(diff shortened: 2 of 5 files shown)


allglc (changes requested):
Thank you for submitting this PR!
I made a minor modification to ensure formatting is correct and passes the check in the CI.

However, the current implementation has two problems:
1. One test does not pass (`test_fit_estimator_pipeline_sample_weight`)
2. Also, the original error message mentioned in issue #798 (`ValueError: Pipeline.fit does not accept the sample_weight parameter[...]`) still appears when I used the provided code to reproduce the issue.

Could you investigate?

allglc:
I modified a bit the code to simplify it. For instance no need to check if the estimator supports sample weigths, it seems reasonable to have an error when the estimator does not support sample weights but is given some.
I also extended to Classification.
Some calibration tests changed the state of sklearn routing, which made the tests fail in the CI but not locally, it was fixed.

The sample weights are given to the last elements of the pipeline which seems reasonable but not completely general, not sure if this is an issue.

LEDazzio01:
Thanks for the improvements, @allglc! The simplifications look great — agreed that letting the estimator raise naturally when it doesn't support `sample_weight` is cleaner than the `signature()` inspection approach. The `config_context` fix for the calibration tests is a good catch too.

Regarding routing weights to only the last pipeline step: I think that's the right default for the vast majority of use cases (preprocessing steps rarely need sample weights). If a more general routing mechanism is ever needed, sklearn's metadata routing API could handle it, but that feels like a separate concern.

Let me know if there's anything else needed on my end to move this forward!

allglc (approved):
Everything seems good now, thanks again @LEDazzio01 for your contribution!

PR #891, opened 2026-04-02
feat: Risk Monitoring

allglc:


mapie/exchangeability_testing/risk_monitoring.py
```diff
@@ -0,0 +1,239 @@
...
+    ----------
+    [1] Aleksandr Podkopaev and Aaditya Ramdas. Tracking the risk of a deployed
+    model and detecting harmful distribution shifts.
+    International Conference on Learning Representations, 2022.
+    """
+
+    def __init__(
+        self,
+        risk: RiskLike,
+        confidence_level: float = 0.95,
```
GBrelurut: Note sure if confidence level is the good choice, type I error rate seems more natural

mapie/exchangeability_testing/risk_monitoring.py
```diff
@@ -0,0 +1,239 @@
...
+            Ground-truth binary labels for the reference data.
+        y_pred : NDArray
+            Predicted binary labels for the reference data.
+
+        Returns
+        -------
+        RiskMonitoring
+            The fitted instance.
+        """
+        if self.threshold is not None:
```
GBrelurut, line 159: Maybe just a warning ? some users may want to update reference risk

mapie/exchangeability_testing/risk_monitoring.py
```diff
@@ -0,0 +1,239 @@
...
+            raise ValueError("Threshold is already computed.")
+
+        reference_risk_sequence = self.risk.get_risk_sequence(y_true, y_pred)
+        if reference_risk_sequence.size == 0:
+            raise ValueError(
+                "Reference risk is undefined because no samples satisfy the risk condition."
+            )
+
+        self.reference_risk_upper_bound = hoeffding_bound(
```
GBrelurut, line 171: When a reference value is passed this is probably the value that must be substituted.

mapie/exchangeability_testing/risk_monitoring.py
```diff
@@ -0,0 +1,239 @@
...
+            reference_risk_sequence, self.delta_reference, bound_side="upper"
```
GBrelurut, line 172: does the bound_side depends in the type of risk ("greater_is_better" ?)

mapie/exchangeability_testing/risk_monitoring.py
```diff
@@ -0,0 +1,239 @@
...
+        self.online_risk_sequence_history = np.concatenate(
+            [self.online_risk_sequence_history, new_risk_sequence]
+        )
+
+        # in the current implementation, the bound is recomputed from scratch with the full history
+        new_risk_lower_bound_sequence = conjugate_mixture_empirical_bernstein_bound(
+            self.online_risk_sequence_history,
+            v_opt=1,
+            alpha=self.delta_online,
+            bound_side="lower",
```
GBrelurut, line 219: does the bound_side depends on the type of risk ?

GBrelurut (changes requested):

mapie/exchangeability_testing/risk_monitoring.py
allglc: this is to be consistent: confidence_level is used everywhere else in MAPIE. Maybe indicate in the docstring how it relates to type I error rate?

mapie/exchangeability_testing/risk_monitoring.py
allglc: No, it depends from if we look at the reference (source) or the online (target) risk. Upper for reference, Lower for online. Of course the risk value should be low when errors are low (if "lower_is_better" the risk value will be set as 1-risk)

mapie/exchangeability_testing/risk_monitoring.py
allglc: ok, setting reference_risk in init is a mistake (instead I prefered setting threshold directly but forgot to remove reference_risk). Fixed!

mapie/exchangeability_testing/risk_monitoring.py
GBrelurut: Final decision : name parameter test_level to set type I error

GBrelurut (approved):

PR #894, opened 2026-04-08
chore: modernize CI

allglc:
- no typing on minimum requirements, just tests. Avoids "obselete" typing errors
- Modernized CI workflow structure by splitting checks into dedicated quality, tests, and non-blocking latest-canary jobs.
- Improved reliability and performance with explicit permissions, workflow concurrency cancellation, job timeout-minutes, fail-fast: false, and pip caching via actions/setup-python.
- Updated test matrix to run minimum-supported dependencies on ubuntu-latest with Python 3.9, plus cross-platform tests on Python 3.13 (ubuntu, windows, macos).
- Added a forward-compatibility canary lane using python-version: "3.x" + check-latest: true and upgraded core scientific dependencies, while keeping it non-blocking (continue-on-error: true).
- Upgraded GitHub Actions references to current tags: actions/checkout@v6.0.2 and actions/setup-python@v6.2.0.
- run tests in parallel (up to 2x faster)
- mkdocs caching of gallery and packages to accelerate execution

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,12 @@ jobs:
     name: Tests (${{ matrix.name }})
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
+    env:
+      # Limit BLAS/OpenMP threads so xdist workers do not oversubscribe CPUs.
+      OMP_NUM_THREADS: "1"
+      OPENBLAS_NUM_THREADS: "1"
+      MKL_NUM_THREADS: "1"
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -1,4 +1,4 @@
-name: Unit tests
+name: CI
```
GBrelurut, line 1: Woh, so obvious, thank you !

GBrelurut (approved):

PR #923, opened 2026-05-18
fix: make optimize_beta actually minimize prediction interval width

### LINKED ISSUE #484: Error when computing multiple prediction intervals in Time Series Tutorial
Hello MAPIE team,

Thanks for this wonderful library. I'm encountering an issue when trying to compute two intervals instead of one in the time series tutorial, but I might be wrong or missing something in my understanding.

### **Describe the bug**
When attempting to compute two intervals instead of one by changing the `alpha` value from `0.05` to `[0.1, 0.05]` in the time series tutorial, a `ValueError` is raised. The error message indicates an issue with evaluating the truth value of an array in the `conformity_scores.py` file.

### **To Reproduce**
1. Go to the time series tutorial provided in the MAPIE documentation: [MAPIE Time Series Tutorial](https://mapie.readthedocs.io/en/latest/examples_regression/4-tutorials/plot_ts-tutorial.html).
2. Replace the original `alpha = 0.05` with `alpha = [0.1, 0.05]` on line 193 of the tutorial script.
3. Run the script.
4. See the error.

### **Expected behavior**
The script should compute and return two prediction intervals without raising any errors.

### **Screenshots**
N/A (the error message is provided below).

### **Desktop:**
- OS: Windows 10
- IDE: PyCharm 2024.1 (Community Edition)
- MAPIE Version: 0.8.6

### **Additional context**
I tried to find out what is causing it and here are some of my findings:

In conformity_scores.py, the get_quantile function (line 213) receives an inconsistent formatting of alpha_np, sometimes receiving a simple NDArray and sometimes receiving an array of arrays.

Error traceback:
```
y_pred_enbpi_npfit, y_pis_enbpi_npfit = mapie_enbpi.predict(
  File "C:\Users\pzaho\anaconda3\envs\dynamo\lib\site-packages\mapie\regression\time_series_regression.py", line 459, in predict
    return super().predict(
  File "C:\Users\pzaho\anaconda3\envs\dynamo\lib\site-packages\mapie\regression\regression.py", line 645, in predict
    self.conformity_score_function_.get_bounds(
  File "C:\Users\pzaho\anaconda3\envs\dynamo\lib\site-packages\mapie\conformity_scores\conformity_scores.py", line 454, in get_bounds
    quantile_low = self.get_quantile(
  File "C:\Users\pzaho\anaconda3\envs\dynamo\lib\site-packages\mapie\conformity_scores\conformity_scores.py", line 272, in get_quantile
    quantile = signed * np.column_stack([
  File "C:\Users\pzaho\anaconda3\envs\dynamo\lib\site-packages\mapie\conformity_scores\conformity_scores.py", line 276, in <listcomp>
    ) if not (unbounded and _alpha >= 1) else np.inf * np.ones(n_ref)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

### LINKED ISSUE #588: Parameter optimize_beta is not working
After investigating issue #484, I realized our implementation of the `optimize_beta` option is incorrect.

Code to fix:
 - In function `mapie.conformity_scores.regression.BaseRegressionScore.get_bounds` (the only place where beta optimization is performed), we pass a reshaped version of conformity scores to `_beta_optimize`, and we probably should not.
```python
        if optimize_beta:
            beta_np = self._beta_optimize(
                alpha_np,
                conformity_scores.reshape(1, -1),
                conformity_scores.reshape(1, -1),
            )
```
 - The implementation of `_beta_optimize` seems OK-ish (see below). However, we get `conformity_scores.reshape(1, -1)` for parameters `lower_bounds` and `upper_bounds`, and  the reshape always produces arrays of shape (1, n), so the value of `len(lower_bounds)` is always 1.
 - In `_beta_optimize`, there may be an issue with the type of `_alpha` in the enumerate.
 - In `_beta_optimize`, we distinguish `lower_bounds` and `upper_bounds` even though this value is the same, so it could be simplified.

After fixing this, we should:
 - Implement tests to make sure `optimize_beta` works as expected
 - Verify example notebooks that are using this feature (are the results still consistent?)

The MAPIE core team is currently in discussion regarding the priority of this fix VS other ongoing developments. Contributions are welcome :)

immu4989:
# Description

`_beta_optimize` has been a silent no-op since MAPIE v0.4 (per @Valentin-Laurent's
analysis in #588). The caller in `BaseRegressionScore.get_bounds` reshaped
`conformity_scores` to `(1, n)` before passing it to `_beta_optimize`, which
made `len(lower_bounds) == 1` inside the function and collapsed the β-search
grid (`np.linspace(α/(n+1), α, num=n)`) to a single candidate. The output
shape `(n_samples, n_alpha)` also mismatched the non-optimize branch
(`alpha_np / 2`, shape `(n_alpha,)`), producing the wrong prediction-interval
shape when `optimize_beta=True` was used with multiple confidence levels,
the symptom reported in #484.

This PR fixes the four defects enumerated in #588:

1. **Reshape collapsing the search grid** : drop `.reshape(1, -1)` in
   `get_bounds`; pass `conformity_scores` directly as 1D.
2. **`len(lower_bounds)` always 1** : auto-resolved once the reshape is gone.
3. **`_alpha` dtype handling in `enumerate`** — cast `float(_alpha)` at the
   top of the loop to neutralize numpy 2.x dtype-promotion surprises when
   `alpha_np` is `float32`.
4. **Wrong PI shape with a list of alphas** : `_beta_optimize` now returns
   shape `(n_alpha,)`, matching the non-optimize branch.

The two argument signature for `lower_bounds` / `upper_bounds` is kept
unchanged , per @Valentin-Laurent's clarification in the issue thread,
this is intentional for CQR (conformalized quantile regression), where
the lower and upper residual distributions can differ from one another.

Also removed the `logging.warning` (and `import logging`) that flagged the
broken state.

Fixes #588.
Also resolves #484 (`ValueError` on multiple prediction intervals in the
time series tutorial with `optimize_beta=True`).

## Type of change

- Bug fix (non breaking change which fixes an issue)

# How Has This Been Tested?

Full local CI passes: `make lint && make format && make type-check && make coverage`
, **2522 tests passing, 100% coverage** in 1m 43s.

- [x] **New focused unit tests** in `test_conformity_scores_bounds.py`:
  - `test_beta_optimize_shape_and_range` (parameterized over single and
    multi alpha): pins output shape `(n_alpha,)`, dtype `float64`, no NaNs,
    and that each β lies in the search grid range `[α/(n+1), α]`.
  - `test_beta_optimize_handles_float32_alpha`: pins the dtype-cast fix.
- [x] **`test_non_regression.py`**: removed the
  `if not minimize_interval_width:` guard around the shape assertion.
  The assertion `pred_intervals.shape == (X_test_length, 2, n_confidence_level)`
  now fires for the multi-alpha + `optimize_beta=True` test cases , the
  exact surface that reproduced #484.
- [x] **`test_time_series_regression.py`**: the two `_wopt` strategies in
  the `WIDTHS` table previously had values identical to their non-`_wopt`
  counterparts (3.89 == 3.89, 3.85 == 3.85) , a direct fingerprint of the
  no-op. After the fix: optimized-mean drops to 3.84 (narrower, as
  expected); optimized-median moves to 3.89 (β picked away from α/2 by
  the optimizer); coverage held at 0.956 in both cases within `rtol=1e-2`.
  The inline expected `width_mean` in
  `test_interval_prediction_with_beta_optimize` was similarly recomputed
  (3.67 → 3.59).

Reproduce:
uv sync --python 3.12 --extra dev --extra docs --extra notebooks
make lint && make format && make type-check && make coverage

# Checklist

## Guidelines

## Quality Checks
- [ ] When updating documentation: doc builds successfully and without warnings: `make doc`
- [ ] When updating documentation: code examples in doc run successfully: `make doctest`

## Contribution Documentation

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -54,5 +54,6 @@
 * Hannes Körner <HannesMK>
 * L. Elaine Dazzio <elaine.dazzio@gmail.com>
 *- Sarthak Bhide (@Tamish03) <neetasarthak26@gmail.com>
+* Imran Ahamed <immu4989@gmail.com>
 
 To be continued ...
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,7 @@
 * Simplify internal `sample_weight` handling in quantile regression module: `sample_weight` now flows through `fit_params` instead of being passed as a separate argument through the call chain. No public API changes. (issue #753)
 * Remove `_prepare_fit_params_and_sample_weight` utility (no longer needed after regression, classification, and quantile regression refactors). (issue #753)
 * Add notebook kernel restart warning for Kaggle/Jupyter/Colab users after installation or version changes. (issue #916)
+* Fix `optimize_beta` in regression conformity scores so prediction interval width minimization actually optimizes β (was previously a no-op due to a shape-collapsing reshape); also resolves incorrect prediction interval shape when used with multiple confidence levels. (issues #588, #484)
 
 ## 1.4.0 (2026-04-30)
 
--- a/mapie/conformity_scores/regression.py
+++ b/mapie/conformity_scores/regression.py
@@ -1,4 +1,3 @@
-import logging
 from abc import ABCMeta, abstractmethod
 from typing import Tuple, cast
 
@@ -186,7 +185,7 @@ class BaseRegressionScore(BaseConformityScore, metaclass=ABCMeta):
 
         Parameters
         ----------
-        alpha_np: NDArray
+        alpha_np: NDArray of shape (n_alpha,)
             The quantiles to compute.
 
         upper_bounds: NDArray of shape (n_samples,)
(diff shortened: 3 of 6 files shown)


GBrelurut (approved):
Thank you for fixing the issue !

PR #914, opened 2026-05-01
refactor: simplify sample_weight handling in quantile regression module

LEDazzio01:
## Summary

Completes the three-phase simplification of internal `sample_weight` handling (issue #753):

| Phase | Module | PR | Status |
|-------|--------|----|--------|
| 1 | Regression | #902 | ✅ Merged |
| 2 | Classification | #908 | ✅ Merged |
| **3** | **Quantile Regression** | **This PR** | 🔄 Open |

### Changes

- `_MapieQuantileRegressor.fit()`: `sample_weight` is now extracted from `**fit_params` instead of being a separate parameter
- `_MapieQuantileRegressor._fit_estimators()`: same — extracted from `**fit_params`
- `ConformalizedQuantileRegressor.fit()`: uses `_prepare_params()` instead of `_prepare_fit_params_and_sample_weight()`
- **Remove `_prepare_fit_params_and_sample_weight`** from `utils.py` — zero callers remaining after all three phases
- Updated tests and `HISTORY.md`

### Design Notes

- **No public API changes.** The public `ConformalizedQuantileRegressor` still accepts `sample_weight` inside `fit_params` as before.
- `_prepare_train_calib` and `_train_calib_split` keep `sample_weight` as an explicit parameter since they need to split it alongside X/y (analogous to `EnsembleRegressor` and `EnsembleClassifier`).
- Both `# type: ignore[override]` markers are retained — the method signatures still intentionally diverge from the parent class.

### Testing

- All 114 quantile regression tests pass
- All 512 common + regression + utils tests pass  
- Lint (`ruff check`), format (`ruff format`), and type check (`mypy`) all clean

Fixes #753 (final phase)

PR #902, opened 2026-04-18
refactor: simplify sample_weight handling in regression module

LEDazzio01:
Closes #753 (regression module — Phase 1 of 3)

## Summary

Simplifies the internal flow of `sample_weight` in the regression module by keeping it inside `fit_params` throughout the call chain, instead of extracting it via `_prepare_fit_params_and_sample_weight` and threading it as a separate argument.

**Net result: 29 insertions, 47 deletions → 18 fewer lines of code.**

### `mapie/regression/regression.py`
- `SplitConformalRegressor.fit`, `CrossConformalRegressor.fit_conformalize`, and `JackknifeAfterBootstrapRegressor.fit_conformalize` now use `_prepare_params` instead of `_prepare_fit_params_and_sample_weight`, keeping `sample_weight` in `fit_params`
- `_MapieRegressor.fit`, `init_fit`, `fit_estimator`, and `conformalize` no longer take `sample_weight` as a separate parameter
- `_check_fit_parameters` extracts `sample_weight` from `self._fit_params` for null-weight validation via `_check_null_weight`, then puts it back
- Python's keyword argument matching routes `sample_weight` from `**fit_params` to the existing explicit parameters in `EnsembleRegressor` and `_fit_estimator` — **no changes needed in those files**

### `mapie/regression/quantile_regression.py`
- Added `# type: ignore[override]` to `_MapieQuantileRegressor.fit` and `conformalize` since the parent signature changed (will be updated in Phase 3)

### `mapie/tests/test_common.py`
- `test_default_sample_weight`: narrowed to `_MapieClassifier` only, since `_MapieRegressor.fit` no longer has an explicit `sample_weight` parameter

### `HISTORY.md`
- Added changelog entry

## Scope

This PR covers **Phase 1 (regression module only)** of the refactoring described in #753. Classification (Phase 2) and quantile regression (Phase 3) can follow in separate PRs if this approach looks good.

## Verification

| Check | Result |
|-------|--------|
| Full test suite (2361 tests) | ✅ All passed |
| Lint (`ruff check`) | ✅ Passed |
| Format (`ruff format`) | ✅ Passed |
| Type-check (`mypy`) | ✅ Passed |

## No public API changes

This is a pure internal refactoring. The v1 public classes (`SplitConformalRegressor`, `CrossConformalRegressor`, etc.) continue to accept `sample_weight` inside `fit_params` exactly as before.

Issue #753, opened 2025-09-01
Simplify sample_weights handling (internal refactoring)


Valentin-Laurent:
Release v1 has simplified the way sample_weights are passed as arguments (directly in `fit_params`).

We can simplify the fit methods of classes `_MapieRegressor`, `_MapieQuantileRegressor` and `_MapieClassifier` to include `sample_weight` directly in `fit_params` as well, thus removing the need of the specific `_prepare_fit_params_and_sample_weight` function and its usage in `ConformalizedQuantileRegressor`, `CrossConformalRegressor`, `JackknifeAfterBootstrapRegressor` and `CrossConformalClassifier`.

We can also most probably remove `_fit_estimator` (even for calibration), and replace its usage by a simple `estimator.fit(X, y, **fit_params)` (I believe the checks performed in this function are too specific and unnecessary once sample_weights are included in `fit_params`

LEDazzio01:
I've opened #902 to address Phase 1 of this issue — the regression module (`_MapieRegressor`, `SplitConformalRegressor`, `CrossConformalRegressor`, `JackknifeAfterBootstrapRegressor`).

The approach keeps `sample_weight` inside `fit_params` throughout the call chain and leverages Python's keyword argument matching to route it to the existing explicit `sample_weight` parameters in `EnsembleRegressor` and `_fit_estimator` — so those files needed zero changes.

Classification (Phase 2) and quantile regression (Phase 3) can follow in separate PRs if the approach looks good.

GBrelurut:
Hello @VishnoiAman777,
I'm really sorry for these months without answer. Thank you very much for your implication and your time, this is not an easy one.
Actually most of your hypotheses hold. My only reservation concerns the reliance on sklearn model fit. Mapie can be used with various kind of models. So this is not an hypothesis we can accept.

GBrelurut:
Hello @LEDazzio01 
Thank you for contributing. I think this is the way to go.

PR #926, opened 2026-05-20
fix: add defensive validation for sample_weight routing

LEDazzio01:
## Summary

Completes the `sample_weight` refactor series (Phase 1: #902, Phase 2: #908, Phase 3: #914) by adding **runtime enforcement** of the internal contract: `sample_weight` must always flow through `fit_params`, never as a top-level keyword argument.

Also fixes a **bug in `TimeSeriesRegressor`** where `sample_weight` passed as a top-level kwarg was silently ignored.

## Motivation

After the Phase 1-3 refactor, `_MapieRegressor.fit()` and `_MapieClassifier.fit()` use `**kwargs`, so a caller could write:

```python
mapie.fit(X, y, sample_weight=weights)  # silently swallowed — never applied!
```

The weight ends up in `kwargs` but is never extracted or used. This PR catches that mistake with a clear `TypeError` and migration message.

### Bug: `TimeSeriesRegressor` tests silently broken

`TimeSeriesRegressor` inherits `_MapieRegressor.fit(**kwargs)`. The test `test_results_with_constant_sample_weights` (lines 226-228) passed `sample_weight` as a top-level argument — the weights were **never applied**, so the test was not actually verifying weighted behavior.

## Changes

- **`mapie/utils.py`**: Add `_check_deprecated_sample_weight_kwarg()` utility
- **`mapie/regression/regression.py`**: Add validation in `_MapieRegressor.init_fit()` after `fit_params` is popped from kwargs
- **`mapie/classification.py`**: Add validation in `_MapieClassifier.fit()` after `fit_params` is popped from kwargs
- **`mapie/tests/test_time_series_regression.py`**: Fix `sample_weight` routing to use `fit_params={"sample_weight": ...}` + add negative test
- **`mapie/tests/test_regression.py`**: Add negative test
- **`mapie/tests/test_classification.py`**: Add negative test
- **`HISTORY.md`**: Changelog entry

No public API changes. Issue #753.

## LLM Usage

- **LLM used**: Google Gemini (via Antigravity)
- **Purpose**: Codebase analysis to identify the silent failure path in `TimeSeriesRegressor`, and code generation for the defensive validation utility and tests

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,7 @@
 * Remove `_prepare_fit_params_and_sample_weight` utility (no longer needed after regression, classification, and quantile regression refactors). (issue #753)
 * Add notebook kernel restart warning for Kaggle/Jupyter/Colab users after installation or version changes. (issue #916)
 * Fix `optimize_beta` in regression conformity scores so prediction interval width minimization actually optimizes β (was previously a no-op due to a shape-collapsing reshape); also resolves incorrect prediction interval shape when used with multiple confidence levels. (issues #588, #484)
+* Add defensive validation: `_MapieRegressor` and `_MapieClassifier` now raise `TypeError` when `sample_weight` is passed as a top-level keyword argument instead of inside `fit_params`. Previously, top-level `sample_weight` was silently ignored. Also fix `TimeSeriesRegressor` tests that were affected by the same silent-ignore bug.
 
 ## 1.4.0 (2026-04-30)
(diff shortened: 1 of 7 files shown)


GBrelurut (approved):
Thank you @LEDazzio01  for adding the defensive line.

The copilot point on deprecation because of the signature change is also a good point. We will adress this before releasing the patch.

PR #931, opened 2026-05-27
fix: allow refitting CrossConformalRegressor via fit_conformalize and add reset()

### LINKED ISSUE #710: ValueError on Multiple Calls to fit_conformalize in CrossConformalRegressor
This is not a bug, but a question regarding the following code:

```python
from sklearn.datasets import make_regression
from mapie.regression import CrossConformalRegressor
from sklearn.linear_model import Ridge

model = CrossConformalRegressor(estimator=Ridge())

X, y = make_regression()
model.fit_conformalize(X, y)

another_X, another_y = make_regression()
model.fit_conformalize(another_X, another_y)
```

When executed, this code raises a `ValueError: fit_conformalize method already called. MAPIE does not currently support calling fit_conformalize several times.`

I couldn't find any information about this behavior in the documentation. It surprised me as it differs from the behavior in versions prior to v1 and also from scikit-learn, which allows the object to be reused. I'm curious about the reasoning behind this implementation. Additionally, I'm wondering if, for my needs, I have no choice but to create a new instance of the Mapie regressor each time I use it on a new dataset.

immu4989:
# Description

Implements the design @GBrelurut confirmed in
[#710 (comment)](https://github.com/scikit-learn-contrib/MAPIE/issues/710#issuecomment-4541586434):

> "I love the idea of the `.reset()` method to have an explicit action.
> I would have it used in the fit() method with a warning."
> "I think it is best to have a narrow scope on CrossConformalRegressor and
> propagate later to other objects."

Concretely:

- **Adds a public `reset()` method** on `CrossConformalRegressor` that discards
  the conformity scores computed during a prior `fit_conformalize`, returning
  the instance to its pre-fit state (mirrors sklearn's fluent-API convention by
  returning `self`).
- **Replaces the `_raise_error_if_method_already_called` guard inside
  `fit_conformalize`**: on an already-fitted instance, the method now emits a
  `UserWarning` and internally calls `self.reset()` before proceeding with the
  new fit. Users who want surprise-free behavior can call `.reset()` explicitly
  to suppress the warning.

The underlying calibration-leakage concern that originally motivated the guard
(per @GBrelurut: *"the intention was to avoid cases where a model could be
trained anew and conformity scores were not updated"*) is preserved through the
internal state-reset, not via the hard error.

**Scope is intentionally narrow to `CrossConformalRegressor`** per @GBrelurut's
guidance. The same pattern can be propagated to `JackknifeAfterBootstrapRegressor`,
`CrossConformalClassifier`, `SplitConformalRegressor`,
`ConformalizedQuantileRegressor`, and `SplitConformalClassifier` in follow-up PRs.

Fixes #710.

## Type of change

- This change requires a documentation update (docstring updated on
  `fit_conformalize`; new docstring on `reset()`)

# How Has This Been Tested?

Full local CI passes: `make lint && make format && make type-check && make coverage`
— **2525 tests passing (+3 from new tests), 100% coverage** in 1m 37s.

- [x] `test_common.py::TestWrongMethodsOrderRaisesErrorForCrossTechniques::test_wrong_methods_order`:
  the existing test now branches — for `CrossConformalRegressor` it asserts a
  `UserWarning` is emitted on the second `fit_conformalize` call; for
  `JackknifeAfterBootstrapRegressor` and `CrossConformalClassifier` it still
  asserts the existing `ValueError` (unchanged behavior — those classes are
  out of scope per #710's narrow scope).
- [x] **New** `test_common.py::TestCrossConformalRegressorReset` with three
  focused tests:
  - `test_reset_clears_state`: `.reset()` returns `self`, clears
    `is_fitted_and_conformalized`, and resets `_predict_params` to `{}`.
  - `test_explicit_reset_then_refit_does_not_warn`: calling `.reset()` before
    `fit_conformalize` suppresses the `UserWarning`
    (verified with `warnings.simplefilter("error", UserWarning)`).
  - `test_refit_produces_calibration_from_second_dataset`: fit on `y * 0.5`,
    then refit on `y * 50.0`; verifies the resulting prediction intervals are
    bit-identical to a freshly-constructed instance fit only on `y * 50.0`.
    This proves there is no calibration leakage from the first fit.

Reproduce:
uv sync --python 3.12 --extra dev --extra docs --extra notebooks
make lint && make format && make type-check && make coverage

# Checklist

## Guidelines

## Quality Checks

## Contribution Documentation
- [x] I have updated the [HISTORY.md](https://github.com/scikit-learn-contrib/MAPIE/blob/master/HISTORY.md) and [AUTHORS.md](https://github.com/scikit-learn-contrib/MAPIE/blob/master/AUTHORS.md) files (HISTORY.md updated this PR; AUTHORS.md already lists me from #923).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,7 @@
 * Add notebook kernel restart warning for Kaggle/Jupyter/Colab users after installation or version changes. (issue #916)
 * Fix `optimize_beta` in regression conformity scores so prediction interval width minimization actually optimizes β (was previously a no-op due to a shape-collapsing reshape); also resolves incorrect prediction interval shape when used with multiple confidence levels. (issues #588, #484)
 * Add defensive validation: `_MapieRegressor` and `_MapieClassifier` now raise `TypeError` when `sample_weight` is passed as a top-level keyword argument instead of inside `fit_params`. Previously, top-level `sample_weight` was silently ignored. Also fix `TimeSeriesRegressor` tests that were affected by the same silent-ignore bug.
+* Add `reset()` method on `CrossConformalRegressor` and allow refitting via `fit_conformalize` (now emits a `UserWarning` and discards prior conformity scores instead of raising). Same pattern can be propagated to other conformal classes in follow-up PRs. (issue #710)
 
 ## 1.4.0 (2026-04-30)
 
--- a/mapie/regression/regression.py
+++ b/mapie/regression/regression.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import Any, Iterable, Optional, Tuple, Union, cast
 
 import numpy as np
@@ -427,6 +428,20 @@ class CrossConformalRegressor:
 
         self._predict_params: dict = {}
 
+    def reset(self) -> CrossConformalRegressor:
+        """
+        Discard previously computed conformity scores so that
+        `fit_conformalize` can be called again with new data.
+
+        Returns
+        -------
+        Self
+            This CrossConformalRegressor instance, reset to its pre-fit state.
+        """
+        self.is_fitted_and_conformalized = False
+        self._predict_params = {}
+        return self
+
     def fit_conformalize(
         self,
         X: ArrayLike,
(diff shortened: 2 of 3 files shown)


GBrelurut (approved):
Thank you for the fix !

PR #934, opened 2026-05-28
feat: add classwise flag to expected_calibration_error

### LINKED ISSUE #277: Metrics limited to a single Expected Calibration Error (ECE) formula
**Describe the bug**
MAPIE offers only one way to calculate the ECE, but there are different estimators of the ECE in the literature (reference: https://arxiv.org/abs/2109.03480).

**To Reproduce**
When using the `expected_calibration_error(y_true: ArrayLike, y_scores: ArrayLike)` function, the computed estimator of ECE is the _class-specific ECE associated to class c_ (see https://arxiv.org/abs/2109.03480 in page 4). 

To use this function correctly, the array `y_true` must be the boolean array indicating whether the example at index `i` is associated with the class considered (named `c`). So for any example at the index `i`, `y_true[i] = 1` if this example is related to the considered class `c`.

**Expected behavior**
It will be interesting to implement other ECE estimators, such as the confidence ECE presented in https://arxiv.org/abs/1706.04599.

**Additional context**
Links for references:
- Guo, C., Pleiss, G., Sun, Y., & Weinberger, K. Q. (2017, July). On calibration of modern neural networks. In International conference on machine learning (pp. 1321-1330). PMLR. (https://arxiv.org/abs/1706.04599)
- Posocco, N., & Bonnefoy, A. (2021). Estimating expected calibration errors. In Artificial Neural Networks and Machine Learning–ICANN 2021: 30th International Conference on Artificial Neural Networks, Bratislava, Slovakia, September 14–17, 2021, Proceedings, Part IV 30 (pp. 139-150). Springer International Publishing. (https://arxiv.org/abs/2109.03480)

Code-ve:
# Description

This PR adds support for classwise Expected Calibration Error (ECE) in `expected_calibration_error` through a new `classwise` parameter.

The default behavior remains unchanged. When `classwise=False`, the function computes the existing confidence-based ECE. When `classwise=True`, ECE is computed independently for each class and averaged across classes.

The PR also adds tests covering the new functionality and updates the related documentation.

Fixes #277

## Type of change

* New feature (non-breaking change which adds functionality)
* This change requires a documentation update

# How Has This Been Tested?

The following tests were added:

* `test_ece_classwise_basic`
* `test_ece_classwise_vs_default`
* `test_ece_classwise_1d_raises_error`
* `test_ece_classwise_perfect_calibration`

In addition, the full CI suite passes successfully on Linux, Windows, macOS, and minimum-supported dependency configurations.

# Checklist

## Guidelines

* [x] I have read the contributing guidelines.
* [x] I have read and followed the testing guidelines.

## Quality Checks

* [x] Linting passes successfully.
* [x] Type checking passes successfully.
* [x] Unit tests pass successfully.
* [ ] Coverage is 100%.
* [x] Documentation builds successfully.
* [ ] Documentation doctests were run.

## Contribution Documentation

* [x] I have updated the relevant project documentation files.

## LLM Usage

* [x] I used a Large Language Model (LLM) for this contribution.
* [x] I carefully reviewed and verified all LLM-assisted suggestions before applying them.

**LLM used:** ChatGPT

**Purpose:** Assistance with reviewing implementation details and refining documentation.


immu4989:
Hi @Code-ve — thanks for picking up #277, the classwise-ECE addition is a useful
extension. A few substantive points and a couple of nits, in rough order of
importance.

## Substantive

**1. The `y_true` docstring is now ambiguous between the two modes.**

After this change, `y_true` means:
- `classwise=False`: binary 0/1 indicator of whether `y == y_pred`.
- `classwise=True`: raw integer class label in `[0, n_classes-1]`.

The new docstring `"Whether the prediction is correct (y == y_pred), encoded as 0
or 1"` is right for the default case but actively wrong for the classwise case
(you're passing raw class labels, not correctness indicators). A user reading
the docstring while using `classwise=True` would pass the wrong array and get
silently wrong numbers.

Two ways to fix:

- Document both interpretations in the same `y_true` description, OR
- Split into two functions: keep the existing `expected_calibration_error` as
  the confidence-ECE entry point, add a new `classwise_ece` or
  `expected_calibration_error_classwise`. The Nixon et al. paper treats these
  as distinct metrics with different inputs, which argues for distinct
  callsites.

I lean toward splitting (clearer API, no input-meaning-changes-based-on-flag
foot-gun), but extending the docstring is acceptable if the maintainers prefer
a single entry point.

**2. Missing reciprocal validation between `y_true` range and `y_scores.shape[1]`.**

Currently:
```python
n_classes = y_scores.shape[1]
for c in range(n_classes):
    y_true_c = np.array(cast(NDArray, y_true) == c, dtype=int)
```
Two failure modes that produce silently wrong numbers:

- If y_true contains class 5 but y_scores.shape[1] == 3, class 5 is silently ignored — those samples become "negative" in every per-class binary problem.
- If y_scores.shape[1] == 5 but y_true only contains classes [0, 1, 2], classes 3 and 4 contribute a degenerate ECE (all-zero y_true_c against potentially non-trivial scores).

Worth a validation check (y_true.max() < n_classes and possibly the converse)
or at minimum documenting the assumption. The existing tests happen to keep
ranges aligned by construction (randint(0, 3, 100) paired with
dirichlet(np.ones(3), ...)), so this gap isn't caught.

**3. Citation should be double-checked.**

The PR cites arxiv 1910.12656 as Nixon et al. "Measuring Calibration in Deep
Learning" (CVPR Workshops 2019). My recollection is that the Nixon et al. paper
is at arxiv 1904.01685, and that 1910.12656 is a different paper. Could be wrong
— worth confirming the arxiv ID matches the title/authors.

**4. Averaging strategy should be explicit in the docstring.**

The implementation does uniform averaging: ece / n_classes. Nixon et al's
classwise-ECE definition uses uniform averaging; some other definitions in the
literature weight by class prior. A single docstring sentence noting "uniform
mean across classes" prevents readers from having to infer it from code.

**Minor**

- np.size(y_scores.shape) != 2: idiomatic Python is y_scores.ndim != 2. The current expression works but reads as unusual.
- ece = float(0.0): the float() wrap is redundant — 0.0 is already a float.
- Redundant casts: y_true = cast(NDArray, y_true) is followed by cast(NDArray, y_true) == c on the next line. The second cast is unnecessary.
- One worthwhile additional test: a 2-class case with 2D y_scores to verify the classwise path collapses correctly to confidence-ECE (up to the symmetry that y_scores[:, 0] and y_scores[:, 1] are complementary).

**On test_ece_classwise_perfect_calibration**
The test asserts scr < 0.05. Mathematically, one-hot predictions on correct labels should give exactly 0 (modulo binning artifacts). What value does this actually produce locally? If it's < 1e-10, a tighter bound (< 1e-6) would
catch off-by-one regressions in the per-class accumulator that the current 0.05
ceiling wouldn't.

**Question for maintainers**
Would you prefer this to land as classwise=True on the existing function (current PR), or as a separate classwise_ece function? The y_true semantics differing between the two modes is the design question worth resolving here.

Code-ve:
Hi, thanks for taking the time to review this and for the detailed feedback!

You raise several good points, especially around the differing `y_true` semantics between the two modes. I agree that this is the main API-design question here, and I'll take a closer look at whether a separate `classwise_ece` entry point would be clearer than overloading the existing function with a flag. In the meantime, I'll update the documentation so the expected inputs are unambiguous.

I'll also add validation to ensure the class labels in `y_true` are consistent with `y_scores.shape[1]`, revisit the citation to confirm the correct reference, and make the uniform averaging strategy explicit in the docstring.

The implementation nits are helpful as well — I'll clean those up and tighten the perfect-calibration test if the observed value is effectively zero. I'll also look into adding a test covering the 2-class case you suggested.

Thanks again for the thorough review. I'll work through these points and push an update shortly.

allglc:
Hello @immu4989,
Thank you for participating in this conversation. However, I believe that most of your comment is generated by an LLM. If that is the case, please disclose it. Similarly to probably many open-source project maintainers, I prefer to know if I'm talking to a human.

allglc:
Hello @Code-ve ,
Thank you for this contribution. I will review it shortly. In the meantime, please fill the PR template.

mapie/metrics/calibration.py
```diff
@@ -107,38 +107,84 @@ def expected_calibration_error(
...
+
+    When ``classwise=True``, computes the classwise-ECE [2]: the ECE
+    is computed independently for each class and then averaged.
 
     [1] Naeini, Mahdi Pakdaman, Gregory Cooper, and Milos Hauskrecht.
     "Obtaining well calibrated probabilities using bayesian binning."
     Twenty-Ninth AAAI Conference on Artificial Intelligence. 2015.
+    https://doi.org/10.1609/aaai.v29i1.9602
+
+    [2] Nixon, Jeremy, et al.
```
allglc: This is not the correct reference for classwise. This one is https://arxiv.org/abs/1910.12656

mapie/metrics/calibration.py
```diff
@@ -107,38 +107,84 @@ def expected_calibration_error(
...
+    "Measuring Calibration in Deep Learning."
+    CVPR Workshops. 2019.
+    https://arxiv.org/abs/1904.01685
 
     Parameters
     ----------
     y_true: ArrayLike of shape (n_samples,)
-        The target values for the calibrator.
+    When ``classwise=False``, a binary indicator of whether the
```
allglc: The docstring should warn about the misleading variable name `y_true`.

mapie/metrics/calibration.py
```diff
@@ -107,38 +107,84 @@ def expected_calibration_error(
...
+    prediction is correct (``y == y_pred``), encoded as 0 or 1.
+
+    When ``classwise=True``, the true class labels encoded as
```
allglc: `y_true` should not change according to the classwise case or standard case. To keep compatibility and at the cost of a misleading name, `y_true` should in both cases represent the prediction "correctness".

mapie/metrics/calibration.py
```diff
@@ -107,38 +107,84 @@ def expected_calibration_error(
...
 
     [1] Naeini, Mahdi Pakdaman, Gregory Cooper, and Milos Hauskrecht.
     "Obtaining well calibrated probabilities using bayesian binning."
     Twenty-Ninth AAAI Conference on Artificial Intelligence. 2015.
+    https://doi.org/10.1609/aaai.v29i1.9602
+
+    [2] Nixon, Jeremy, et al.
+    "Measuring Calibration in Deep Learning."
+    CVPR Workshops. 2019.
+    https://arxiv.org/abs/1904.01685
```
allglc: It would be nice to add a doctest to clarify what is `y_true`. Typically computed with some sort of `y_pred == label`

allglc (changes requested):

Code-ve:
Thank you for the review. I've updated the PR description to follow the project template. I'm reviewing the comments regarding the y_true semantics and will revisit the documentation and implementation accordingly before making further changes.

mapie/metrics/calibration.py
Code-ve: Thanks for the clarification. I want to make sure I understand the intended API correctly.

The current implementation computes classwise ECE by constructing one-vs-rest targets for each class, which requires access to the true class labels. In the current draft, `y_true` therefore represents class labels when `classwise=True`.

When you mention that `y_true` should continue to represent prediction correctness in both cases, do you envision a different API for providing the class labels needed for the classwise computation, or would you prefer that the implementation be adapted while keeping the existing signature?

I’d like to align the implementation with the intended design before making further changes.

Code-ve:
@allglc,

I've addressed the reference feedback, resolved the merge conflict, and updated the PR accordingly.

Regarding the remaining comments around `y_true`, I left a question in the review thread because I want to ensure the implementation aligns with the intended API design before making further changes. In the current implementation, classwise ECE requires access to the true class labels, so I wanted to confirm the expected direction before proceeding.

Whenever you have a chance, I'd appreciate any guidance on the intended `y_true` semantics so I can address the remaining feedback correctly.

Thank you for your review and guidance.

mapie/metrics/calibration.py
allglc: The API should not change. Even if it's not the best choice. `y_true` should represent the same thing in both cases. To provide the class labels, you can have a new arguement yes.

Code-ve: Thanks for the clarification. I'll update the implementation so that `y_true` retains its current meaning in both modes and introduce a separate argument for the class labels required by the classwise computation.

Code-ve:
Hi @allglc,

I've addressed the review comments, including the citation update, the `y_true` clarification, the API adjustment using a separate `class_labels` argument, the associated tests, and the doctest example clarifying the `y_true` semantics.

The main CI checks are now passing across all platforms. The only remaining failing check is the Read the Docs build, which appears to be timing out during the documentation build step. Since the implementation, tests, formatting, and coverage checks are all passing, I'm not sure whether this timeout is related to the changes in this PR or a transient documentation-build issue.

Please let me know if there is anything else you'd like me to adjust.

Thank you!

mapie/metrics/calibration.py
allglc: THe paper name and authors should be updated according to the linked paper

allglc (changes requested):

allglc:
@Code-ve thanks for the modifications! The implementation is better now. Just one last comment: the reference link for classwise ECE was updated but not the paper title and authors.

And indeed the documentation issue is not linked to this PR, I'm currently solving the issue.

Code-ve:
Thanks for the clarification. I've updated the paper title and authors to match the referenced paper (arXiv:1910.12656).

allglc (approved):
Great, thanks a lot!

Code-ve:
Thank you for the review and guidance throughout the process. I appreciate the feedback and help in improving the implementation.

PR #944, opened 2026-06-06
refactor: unify quantile computation into _compute_quantile (issue #479)

### LINKED ISSUE #479: Share the quantile function between regression and classification
The function that calculates the quantile for regression and classification is not the same.

Reuse one of them for the other task, so you only have to manage one function.

LEDazzio01:
## Description

Extracts the shared quantile logic from `BaseConformityScore.get_quantile` (regression) and `_compute_quantiles` (classification) into a single `_compute_quantile` function in `mapie/utils.py`.

Closes #479

## Changes

The unified function supports both quantile methods via a `method` parameter:
- **`"lower"`** (regression): `ceil(α*(n+1))/n` correction with `method="lower"`, supports `reversed`, `unbounded`, and NaN-aware computation
- **`"higher"`** (classification): `(n+1)*(1-α)/n` correction with `method="higher"`

Both original functions now delegate to `_compute_quantile`:
- `_compute_quantiles` → thin wrapper adding 3D array support for classification
- `BaseConformityScore.get_quantile` → direct delegation

**No public API changes.**

## Files changed

| File | Change |
|---|---|
| `mapie/utils.py` | Add `_compute_quantile`; rewrite `_compute_quantiles` to delegate |
| `mapie/conformity_scores/interface.py` | `get_quantile` delegates to `_compute_quantile`; remove unused `numpy` import |
| `mapie/tests/test_utils.py` | Add `TestComputeQuantile` class (8 tests covering all branches) |
| `HISTORY.md` | Add changelog entry |

## Verification

- ✅ `make coverage`: 2564 passed, 100.00% branch coverage
- ✅ `make type-check`: no issues
- ✅ `make lint` + `make format`: all clean
- ✅ Numerical equivalence verified (2D + 3D) against old implementation

## AI Disclosure

This PR was developed with AI assistance.


allglc:
Hello,
Thank you for tackling this issue. However I don't really like the way it's done in this PR: we now have thee different functions to compute quantiles (even if most of the work is delegated in `_compute_quantile`. in my opinion this adds complexity while the goal is is to simplify and remove lines of code. The argument `method` which can be `higher` or `lower` is not clear to me.
A better way would be to keep only one clean function to compute quantiles, and use only this entry point everywhere. I think it should be possible. If not, maybe splitting between regression and classification is actually preferable.
What do you think @GBrelurut ?

LEDazzio01:
Hi @allglc, thanks for the review — these are fair points.

I agree that having three functions (`_compute_quantile`, `_compute_quantiles`, `get_quantile`) adds indirection rather than reducing it, and that the `method="lower"` / `method="higher"` parameter is unclear about what it's actually selecting.

I dug into whether the two paths could be collapsed into a single formula, and the answer is **no** — they use fundamentally different mechanisms to achieve finite-sample coverage:

- **Regression**: `⌈α_ref × (n+1)⌉ / n` with `method="lower"` — coverage is ensured by `ceil()` rounding the quantile level up
- **Classification**: `(n+1)(1-α) / n` with `method="higher"` — coverage is ensured by numpy's `"higher"` interpolation picking the next order statistic

These produce different values for every combination of `n` and `α` I tested, so they can't be interchanged without breaking one path's coverage guarantee.

Given that true unification isn't possible, I think the better approach (as you suggested) is to **split into two clearly named functions** in `utils.py`:

```python
def _compute_regression_quantile(
    conformity_scores, alpha_np, axis=0, reverse=False, unbounded=False
) -> NDArray:
    """ceil()-based correction with method='lower'."""

def _compute_classification_quantile(
    conformity_scores, alpha_np, axis=0
) -> NDArray:
    """(n+1)(1-α)/n correction with method='higher'."""
```

This still achieves the goal of #479 (moving the quantile logic out of `interface.py` into `utils.py`), but each function is self-contained with no ambiguous `method` parameter. `get_quantile` becomes a one-line delegate to `_compute_regression_quantile`, and `_compute_quantiles` wraps `_compute_classification_quantile`.

I'll rework the PR along these lines. Let me know if that sounds good, or if you'd prefer a different approach.

allglc:
Hi, this seems reasonable.
However `_compute_quantiles` and `get_quantile` should disappear and the new functions should be directly used. The goal is to have clean code, not wrappers etc.

LEDazzio01:
Thanks @allglc, done! I've reworked the PR to eliminate both `_compute_quantiles` and `get_quantile` entirely. The new structure is:

- **`_compute_regression_quantile()`** — `ceil()`-based correction with `method="lower"` (used by `BaseRegressionScore.get_bounds`)
- **`_compute_classification_quantile()`** — `(n+1)(1-α)/n` correction with `method="higher"` (used by LAC, APS, RAPS, TopK)

All call sites now use these functions directly — no wrappers, no intermediate `method` parameter. Net change is 125 insertions / 198 deletions (fewer lines than before). Full test suite passes (2511/2511).

allglc:
Nice improvements!

I now have two questions:
- Could you please move the new functions into conformity_scores/utils.py? It makes more sense to me as they are specific to conformity scores only.
- `_compute_regression_quantile`  does not change much the original implementation but `_compute_classification_quantile`  changes a lot the original implementation. I would prefer keeping close to the original. Especially that it seems to change behavior.

Thanks!

LEDazzio01:
@allglc Both points addressed:

**1. Moved functions to `conformity_scores/utils.py`**

`_compute_regression_quantile`, `_compute_classification_quantile`, and `_check_alpha_and_last_axis` are now in `mapie/conformity_scores/utils.py` with all import paths updated across `regression.py`, `aps.py`, `lac.py`, `raps.py`, `topk.py`, and the test file.

One thing worth noting: this introduced a circular import (`conformity_scores/__init__.py` → `regression.py` → `conformity_scores/utils.py` → `regression.py`), since `utils.py` also defines `CONFORMITY_SCORES_STRING_MAP` which references the score classes. I resolved this by making those score-class imports lazy (inside the functions that need them). The quantile functions themselves have no dependency on the score classes, so they are defined cleanly at the top of the module.

**2. Reverted `_compute_classification_quantile` to original implementation**

The function now matches the original `_compute_quantiles` logic exactly:
- `np.quantile` instead of `np.nanquantile`
- No `min(..., 1.0)` clamping on the quantile level
- No `axis` parameter (removed)
- Uses `np.stack` and `cast(NDArray, ...)` as before

All tests pass (154 in `test_utils.py`, 98 in conformity score tests, 829 in regression + classification).

allglc:
Sorry I haven't thought about the circular import, in that case I prefer avoiding it and putting the code back to mapie/utils.py

LEDazzio01:
Moved both functions back to `mapie/utils.py` in the latest push (902c523).

**What changed:**
- `_compute_regression_quantile` and `_compute_classification_quantile` now live in `mapie/utils.py` (no circular import risk)
- All call sites import directly from `mapie.utils` — no wrappers or indirection
- Cleaned up unused imports in `conformity_scores/utils.py`

Full test suite passes (2523/2523). Ready for review whenever you have a chance.

allglc (approved):
Great contribution, thanks for taking into account my comments!

PR #950, opened 2026-06-18
fix: Jaber documentation

allglc:


allglc:
- [x] add ref in API for "Standardized non-conformity score" and check name

GBrelurut (approved):

PR #953, opened 2026-06-18
add aucroc_score and auarc_score uncertainty metrics

### LINKED ISSUE #551: Additional Metrics for uncertainty evaluation
Hi all, I recently accross this paper : https://arxiv.org/abs/2305.19187

It introduces two interesting metrics about uncertainty evaluation. 

The two business questions addressed are : 

	1. Is my uncertainty predictive of my errors? It is expected that larger uncertainties correlates with higher error rates
	2. How much errors do I spare if I reject predictions with an uncertainty cut-off? In relation to selective regression/classification/generation, it is expected that my error rate decreases if I delegate high uncertainty cases to humans (or to dustbin).

The corresponding two metrics are quite easy to implement:
	1. The AUCROC(y_wrong, y_uncertainty), where y_wrong is 1 if the prediction is wrong, and y_uncertainty is simply the prediction uncertainty. This directly translate the ability of uncertainties to rank wrongest responses (in expectation).
	2. The AUARC (Area Under the Accuracy Rejection Curve), which is simply the accuracy score as a function of rejection rate, or uncertainty cut-off. 

Beyond the two basic metrics, we could push further the concept to:
	• Precision recall curve
	• "Mondrianized" metrics, with an additional `groups` parameters, allowing to stratify the analysis by group
	• Extensive utilities to plot diagnostic curves within plotly (as much as sklearn does), with additional information (e.g. the curve of a perfect/random model).

I think this would elegantly complete the existing coverage_scores metrics with metrics closer to business considerations. Moreover, these metrics are almost use case agnostic, since the user can quite easily compute y_wrong as a function of y_true and y_pred, and y_uncertainty as a function of y_pis (e.g. y_uncertainty = y_pis.sum(axis=1) for multiclass classification, which is the length of the prediction set).

Happy to discuss further about this !

DresdenGman:
Adds aucroc_score and auarc_score to a new mapie/metrics/uncertainty.py, implementing the uncertainty evaluation metrics from Lin et al. (2023). Both are numpy-in scalar-out and task-agnostic.

Closes #551

### DIFF
--- a/mapie/metrics/uncertainty.py
+++ b/mapie/metrics/uncertainty.py
@@ -90,30 +90,30 @@ def aucroc_score(
     Quantification for Black-box Large Language Models. TMLR.
     https://arxiv.org/abs/2305.19187
     """
-    y_wrong = column_or_1d(np.asarray(y_wrong, dtype=float))
-    y_uncertainty = column_or_1d(np.asarray(y_uncertainty, dtype=float))
+    y_wrong_arr = column_or_1d(np.asarray(y_wrong, dtype=float))
+    y_uncertainty_arr = column_or_1d(np.asarray(y_uncertainty, dtype=float))
 
-    check_consistent_length(y_wrong, y_uncertainty)
+    check_consistent_length(y_wrong_arr, y_uncertainty_arr)
 
-    if np.any(np.isnan(y_wrong)) or np.any(np.isnan(y_uncertainty)):
+    if np.any(np.isnan(y_wrong_arr)) or np.any(np.isnan(y_uncertainty_arr)):
         raise ValueError("y_wrong and y_uncertainty must not contain NaN values.")
-    if np.any(np.isinf(y_wrong)) or np.any(np.isinf(y_uncertainty)):
+    if np.any(np.isinf(y_wrong_arr)) or np.any(np.isinf(y_uncertainty_arr)):
         raise ValueError("y_wrong and y_uncertainty must not contain Inf values.")
 
-    unique_labels = np.unique(y_wrong)
+    unique_labels = np.unique(y_wrong_arr)
     if not np.all(np.isin(unique_labels, [0.0, 1.0])):
         raise ValueError(
             "y_wrong must be a binary array containing only 0 and 1. "
             f"Got unique values: {unique_labels}."
         )
 
-    if np.any(y_uncertainty < 0):
+    if np.any(y_uncertainty_arr < 0):
         raise ValueError("y_uncertainty must contain only non-negative values.")
 
     # sklearn's roc_auc_score handles the AUC computation;
     # we use uncertainty as the score that should rank positives (wrong=1)
     # higher than negatives (wrong=0).
-    return float(roc_auc_score(y_wrong, y_uncertainty))
+    return float(roc_auc_score(y_wrong_arr, y_uncertainty_arr))
 
 
 def auarc_score(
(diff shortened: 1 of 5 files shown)


allglc:
Hello, thank you for this contribution!
However it looks like the doctest for auarc_score fails, could you check the logs in the CI to solve the issue?

DresdenGman:
Fixed — the doctest expected value was wrong (1.0 is not achievable when 50% of predictions are incorrect). Updated it to the correct value and clarified the docstring description accordingly.

allglc:
Thanks, now it's coverage and quality checks that still fail.

DresdenGman:
Fixed — removed the two unused imports that were failing the ruff quality check.

DresdenGman:
Fixed — quality checks, tests, Bandit, and ReadTheDocs now pass. The remaining Sonarqube failure appears to be a fork CI configuration issue due to the missing SONAR_TOKEN.

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+"""
+
+import numpy as np
+from numpy.typing import ArrayLike
+from sklearn.metrics import roc_auc_score
+from sklearn.utils import column_or_1d
+from sklearn.utils.validation import check_consistent_length
+
+
+def aucroc_score(
```
allglc: I would prefer the name `auroc` (without `_score`, also AUCROC comes from the GitHub issue, but I think "AUROC" is the most widely used term, for instance in the referenced paper).

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    y_wrong: ArrayLike,
```
allglc: Again it follows the issue notation, but in my opinion it is better to use "correctness" (1 if prediction is correct) and "confidence" (higher value means more confidence, less uncertainty). It's the most widely used way to compute the metric, to the best of my knowledge (for instance it's like this in the paper)

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    ranks incorrect predictions.
+
+
+    A high score means that samples with larger uncertainty tend
+    to be the ones where the model is wrong — i.e. the uncertainty
+    is *predictive* of errors.
+
+
+    Defined as ``AUCROC(y_wrong, y_uncertainty)`` following
+    Lin et al. (2023).
```
allglc: This reference does not introduce this metric, remove reference

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    --------
+    >>> import numpy as np
+    >>> from mapie.metrics.uncertainty import aucroc_score
+    >>> y_wrong = np.array([0, 0, 1, 1])
+    >>> y_uncertainty = np.array([0.1, 0.2, 0.8, 0.9])
+    >>> aucroc_score(y_wrong, y_uncertainty)
+    1.0
+
+
+    References
```
allglc: remove reference

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    if np.any(y_uncertainty_arr < 0):
+        raise ValueError("y_uncertainty must contain only non-negative values.")
+
+    # sklearn's roc_auc_score handles the AUC computation;
+    # we use uncertainty as the score that should rank positives (wrong=1)
+    # higher than negatives (wrong=0).
+    return float(roc_auc_score(y_wrong_arr, y_uncertainty_arr))
+
+
+def auarc_score(
```
allglc: just `auarc` is enough for the function name,

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    the accuracy on retained samples is recorded. The AUARC is the
+    area under this curve, normalised to [0, 1].
+
+
+    A higher AUARC means that rejecting uncertain predictions leads
+    to a larger accuracy gain — i.e. the uncertainty is actionable
+    for selective prediction / human-in-the-loop workflows.
+
+
+    Defined as ``AUARC`` following Lin et al. (2023).
```
allglc: Lin et al did not introduce the metric. Nadeem et al., (2009) did.

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+
+    # sklearn's roc_auc_score handles the AUC computation;
+    # we use uncertainty as the score that should rank positives (wrong=1)
+    # higher than negatives (wrong=0).
+    return float(roc_auc_score(y_wrong_arr, y_uncertainty_arr))
+
+
+def auarc_score(
+    y_wrong: ArrayLike,
+    y_uncertainty: ArrayLike,
```
allglc: Again, I don't think `y_wrong` and `y_uncertainty` the best user facing API, nor the best way to use the metric. Looking at the metric definiton, what do you think is best?

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    >>> y_uncertainty = np.array([0.9, 0.8, 0.2, 0.1])
+    >>> auarc_score(y_wrong, y_uncertainty)
+    0.7916666666666666
+
+
+    References
+    ----------
+    Lin et al. (2023). Generating with Confidence: Uncertainty
+    Quantification for Black-box Large Language Models. TMLR.
+    https://arxiv.org/abs/2305.19187
```
allglc: not the correct reference, either add or replace it with (Nadeem et al., 2009)

mapie/metrics/uncertainty.py
```diff
@@ -0,0 +1,228 @@
...
+    cumulative_correct_from_tail = np.cumsum(y_correct_sorted[::-1])[::-1]
+
+    retained_counts = np.arange(n, 0, -1, dtype=float)  # n, n-1, ..., 1
+    accuracy_curve = cumulative_correct_from_tail / retained_counts
+
+    # AUARC = mean accuracy across all rejection thresholds (0 rejected,
+    # 1 rejected, ..., n-1 rejected), i.e. when retaining n, n-1, ..., 1
+    # samples respectively. We exclude the point where 0 samples remain.
+    auarc = float(np.mean(accuracy_curve))
+    return auarc
```
allglc: not sure I understand the implementation. It seems the paper authors made it much simpler: https://github.com/zlin7/UQ-NLG/blob/ecaea91741a6c1076e9069b56d32a9929eea8ff9/pipeline/eval_uq.py#L10

allglc (changes requested):
Thank you for making changes to pass the CI. I have comments on the implementation. Please answer them when you have the time.

DresdenGman:
Thanks for the detailed review. I updated the API to use `auroc(correctness, confidence)` and `auarc(correctness, confidence)`, removed the incorrect Lin et al. attribution, and simplified the AUARC implementation around confidence-sorted cumulative accuracy. I also updated the tests accordingly.

allglc:
Could you add the reference for AUARC? You can find it in https://arxiv.org/abs/2305.19187

DresdenGman:
Thanks, I added the Nadeem et al. (2009) reference for AUARC in the `auarc` docstring.

allglc (approved):
Great, thank you for this contribution!

PR #958, opened 2026-07-10
Cross cqr

### LINKED ISSUE #200: [ENHANCEMENT] Cross-CQR
At the moment, conformalized quantile regression has only been implemented with the split-conformal strategy. It would be worth extending it to cross-CQR with the `cv` argument.

GBrelurut:
# Description

This PR introduces `CrossConformalizedQuantileRegressor`, a new estimator that extends conformalized quantile regression to the cross-conformal setting.

Unlike `ConformalizedQuantileRegressor` (which requires a dedicated hold-out conformalization set), this estimator uses cross-validation to fit an ensemble of quantile regressors on different folds and calibrates them in an out-of-fold fashion, allowing the full training set to be used for both fitting and conformalization.

fixes #200 

**Changes included:**

- Added `CrossConformalizedQuantileRegressor` to `mapie/regression/quantile_regression.py`, inheriting from  new `_QuantileConformalizer`.
- Added new Quantile specific conformity scores : `QuantileRegressionScore` and  `AbsoluteQuantileRegressionScore`
- Exposed the new class in `mapie/regression/__init__.py` and registered it in the API documentation (`doc/api/regression.md`).
- Added a standalone gallery example: `examples/regression/2-advanced-analysis/plot_cross_conformalized_quantile_regressor.py`.
- Completed the docstring of `CrossConformalizedQuantileRegressor` (parameters, description).
- Add interface  `_Conformalizer` for future updates of the back-end

## Type of change

- This change requires a documentation update

# How Has This Been Tested?

- [x] New unit tests covering `CrossConformalizedQuantileRegressor` (predict shape, aggregation strategies, conformalization, reset) in `mapie/tests/test_quantile_regression.py`.
- [x] Existing test suite passes: `python -m pytest mapie/tests/test_quantile_regression.py` — exit 0.
- [x] `test_utils.py` failures resolved and confirmed passing.
- [x] Gallery example runs end-to-end with `GradientBoostingRegressor` on a synthetic regression task.

# Checklist

## Guidelines

## Quality Checks

## Contribution Documentation

## LLM Usage

- **LLM used**: Claude Sonnet 4.6 (via GitHub Copilot)
- **Purpose**: code generation, documentation writing, test writing, bug fixing


mapie/conformity_scores/interface.py
```diff
@@ -83,6 +84,77 @@ def get_conformity_scores(self, y: NDArray, y_pred: NDArray, **kwargs) -> NDArra
...
+            could be produced (when alpha_np is greater than or equal to 1.).
+
+            By default `False`.
+
+        Returns
+        -------
+        NDArray of shape (1, n_alpha) or (n_samples, n_alpha)
+            The quantiles of the conformity scores.
+        """
+        alpha_np = np.atleast_1d(np.asarray(alpha_np, dtype=float))
```
allglc: move before function call

mapie/regression/quantile_regression.py
```diff
@@ -26,9 +33,1170 @@
...
 
+from mapie.aggregation_functions import aggregate_all
 from .regression import _MapieRegressor
+from mapie.estimator.regressor import _Conformalizer
+from mapie.conformity_scores import BaseRegressionScore
+
+REGRESSOR_TYPE = Union[RegressorMixin, Pipeline]
+
+
+class QuantileRegressionScore(BaseRegressionScore):
```
allglc: move in conformity_scores/

mapie/regression/quantile_regression.py
```diff
@@ -26,9 +33,1170 @@
...
+    """
+
+    def __init__(self, sym: bool = False, consistency_check: bool = True) -> None:
+        super().__init__(sym=sym, consistency_check=consistency_check)
+
+    def get_signed_conformity_scores(
+        self, y: NDArray[float], y_pred: NDArray[float], **kwargs
+    ) -> NDArray[np.float64]:
+        """
+        Placeholder for `get_conformity_scores`.
```
allglc: remove comment

mapie/regression/quantile_regression.py
```diff
@@ -26,9 +33,1170 @@
...
+
+    _central_estimator: Optional[RegressorMixin]
+    estimators_: dict[str, List[RegressorMixin]]
+    estimator: RegressorMixin
+    quantiles: NDArray[float]
+    _alphas: NDArray[float]
+    score: QuantileRegressionScore
+
+    def __init__(self) -> None:
+        # For testing
```
allglc: "to run tests"

mapie/regression/quantile_regression.py
```diff
@@ -26,9 +33,1170 @@
...
+            )
+
+    def _check_quantile_estimator(
+        self,
+        estimator: Optional[REGRESSOR_TYPE] = None,
+    ) -> REGRESSOR_TYPE:
+        """
+        Perform several checks on the estimator to check if it has
+        all the required specifications to be used with this methodology.
+        The estimators that can be used in _MapieQuantileRegressor need to
```
allglc: rename _MapieQuantileReg

mapie/regression/quantile_regression.py
```diff
@@ -495,7 +1645,8 @@ def _check_alpha(
                     "Invalid confidence_level. Allowed values are between 0.0 and 1.0."
                 )
             else:
-                alpha_np = np.array([alpha / 2, 1 - alpha / 2, 0.5])
+                alpha_values = [alpha / 2, 1 - alpha / 2, 0.5]
+                alpha_np = np.array(alpha_values)
```
allglc, line 1883: revert change

allglc (changes requested):

mapie/regression/quantile_regression.py
```diff
@@ -1,34 +1,1436 @@
...
+        bounds_low, bounds_up = [], []
+        y_pred: Optional[NDArray] = None
+
+        for alpha in self.alpha:
+            level = str(alpha)
+            alpha_np = np.array([alpha], dtype=float)
+            scores = self.conformity_scores[level]
+
+            if not allow_infinite_bounds:
+                n = self.score.get_effective_calibration_samples(scores)
```
shivamlalakiya, line 1271: **Blocking correctness — the asymmetric score cancels the `n >= 1/alpha` guard, and the clipped level is silently accepted.**

This check sees the wrong `n`. `get_effective_calibration_samples` (`mapie/conformity_scores/regression.py:533-535`) counts *entries* and then halves for asymmetric scores:

```python
n: int = np.sum(~np.isnan(scores))
if not self.sym:
    n //= 2
```

`QuantileRegressionScore.get_signed_conformity_scores` returns shape `(2, n)`, so that is `2n // 2 == n`. The halving that protects every other asymmetric score in MAPIE is exactly cancelled, and `alpha` is checked against the full `n` even though `get_bounds` will split it into `beta = alpha/2` per side (`conformity_scores/regression.py:474`).

Below `n = (2-alpha)/alpha` the requested order statistic does not exist, `_compute_regression_quantile` clips the corrected level to 1.0 (`mapie/utils.py:817`), and both bounds fall back to the extreme order statistic. The `unbounded` escape at `utils.py:832` tests `alpha_ref >= 1`, not `alpha_cor > 1`, so `allow_infinite_bounds=True` changes nothing either. No warning is emitted.

<details>
<summary>Repro</summary>

```python
import numpy as np
from sklearn.linear_model import QuantileRegressor

from mapie.conformity_scores import AbsoluteConformityScore
from mapie.conformity_scores.bounds.quantile import QuantileRegressionScore
from mapie.regression import (
    CrossConformalizedQuantileRegressor,
    SplitConformalRegressor,
)

rng = np.random.default_rng(0)
n, alpha = 15, 0.1
X = rng.normal(size=(n, 3))
y = X @ np.array([1.0, -2.0, 0.5]) + rng.normal(size=n)

# 1. what the guard counts, per score layout
print("QuantileRegressionScore (2, 15) ->",
      QuantileRegressionScore().get_effective_calibration_samples(np.zeros((2, n))))
print("AbsoluteConformityScore (15,)   ->",
      AbsoluteConformityScore(sym=False).get_effective_calibration_samples(np.zeros(n)))

# 2. same n, same alpha, same alpha/2 split: the split estimator refuses
try:
    SplitConformalRegressor(
        confidence_level=1 - alpha,
        conformity_score=AbsoluteConformityScore(sym=False),
        prefit=False,
    ).fit(X, y).conformalize(X, y).predict_interval(X)
except ValueError as e:
    print("SplitConformalRegressor:", str(e).replace("\n", " "))

reg = CrossConformalizedQuantileRegressor(
    estimator=QuantileRegressor(solver="highs-ds", alpha=0.0),
    confidence_level=1 - alpha,
    conformity_score=QuantileRegressionScore(),
    method="base", cv=3, random_state=0,
).fit_conformalize(X, y)
_, itv = reg.predict_interval(X[:2])
_, itv_inf = reg.predict_interval(X[:2], allow_infinite_bounds=True)
print("cross-CQR bounds finite:", np.all(np.isfinite(itv)),
      "| allow_infinite_bounds changes nothing:", np.allclose(itv, itv_inf))
```

```
QuantileRegressionScore (2, 15) -> 15
AbsoluteConformityScore (15,)   -> 7
SplitConformalRegressor: Number of samples of the score is too low, 1/confidence_level and 1/(1 - confidence_level) must be lower than the number of samples.
cross-CQR bounds finite: True | allow_infinite_bounds changes nothing: True
```
</details>

Same `n`, same `alpha`, same asymmetric split: the split estimator refuses, this one returns a finite interval. At `n = 15, alpha = 0.1` the lower side asks for order statistic `ceil(0.95 * 16) = 16` out of 15 available. Isolating the conformal step on exchangeable draws, I measured **0.8725 two-sided coverage against a requested 0.90** over 200k draws.

Windows where the guard passes but the order statistic does not exist:

| alpha | guard needs n >= | correctness needs n >= | silent-clip window |
|-------|------------------|------------------------|--------------------|
| 0.05  | 20               | 39                     | 20..38             |
| 0.10  | 10               | 19                     | 10..18             |
| 0.20  | 5                | 9                      | 5..8               |
| 0.30  | 4                | 6                      | 4..5               |

Suggested fix is a four-line override on the new score — see my comment on `mapie/conformity_scores/bounds/quantile.py:47`.

The deeper fix (raise or return `+inf` instead of clipping at `utils.py:817`) is #974 and out of scope here.

mapie/conformity_scores/bounds/quantile.py
```diff
@@ -0,0 +1,102 @@
...
+
+        y_pred: NDArray[float] of shape (n_samples,)
+            Predicted target values.
+
+        Returns
+        -------
+        NDArray[float] of shape (n_samples, 2)
+            Signed conformity scores.
+        """
+        return np.vstack((y - y_pred[0], y - y_pred[1]))
```
shivamlalakiya, line 47: This `(2, n)` layout is what cancels the asymmetric halving in `get_effective_calibration_samples` — see my comment on `mapie/regression/quantile_regression.py:1271` for the failure. The narrow fix belongs here:

```python
    def get_effective_calibration_samples(self, scores: NDArray[np.float64]) -> int:
        # One row per side: the guard must see the per-side count, not the total.
        scores = np.asarray(scores)
        return super().get_effective_calibration_samples(
            scores[0] if scores.ndim > 1 else scores
        )
```

The `ndim` branch is load-bearing, not defensive padding: `AbsoluteQuantileRegressionScore` subclasses this class and overrides `get_conformity_scores` to return 1-D (`quantile.py:102`), so the default score inherits this override and must keep reading the array whole.

Three things about the blast radius, since tightening a guard usually raises the question:

- **The default path is unchanged.** `AbsoluteQuantileRegressionScore` is `sym=True`, so it still reports 15 for `n = 15` and never reaches the clip — for symmetric scores the existing `n >= ceil(1/alpha)` guard is already strictly stronger than the requirement `n >= (1-alpha)/alpha`.
- **Zero released behaviour changes.** `mapie/conformity_scores/bounds/quantile.py` is a new file in this PR and both scores in it are new, so no released configuration can start raising on upgrade. No deprecation cycle needed.
- **The suite stays green.** I applied this on the branch, ran `pytest mapie/tests/test_quantile_regression.py mapie/tests/test_conformity_scores_bounds.py` → `302 passed`, then reverted it. That no test moved is itself informative — see my comment on `mapie/tests/test_quantile_regression.py:2494`.

mapie/regression/quantile_regression.py
```diff
@@ -1,34 +1,1436 @@
...
+
+    confidence_level : float or Iterable[float], default=0.9
+        Target confidence level of the prediction intervals. An iterable requests
+        one interval per level, which `predict_interval` returns along the last axis.
+
+    conformity_score : QuantileRegressionScore or type, default=AbsoluteQuantileRegressionScore
+        Conformity score used to calibrate the quantile estimates. A class is
+        instantiated with its defaults; pass an instance to configure it.
+
+    method : str, default="plus"
```
shivamlalakiya, line 1037: **API / documentation — `method` does not change the conformal step.**

`_QuantileConformalizer._predict` returns bounds with one column per *confidence level*, and `_narrow_bounds_to_alpha` reduces that to one column. So in the `"plus"` branch of `get_bounds`, `distribution_low` is `y_pred_low + s_i` with `y_pred_low` constant along the calibration axis, and `quantile_i(y_pred_low + s_i) == y_pred_low + quantile_i(s_i)` — algebraically the `else` branch. Measured on n=200, printing `bound - raw_quantile_prediction`:

```
method    lower offset            upper offset            spread across test points
base      -0.067776058533         0.067776058533          0.00e+00   0.00e+00
plus      -0.067776058533         0.067776058533          0.00e+00   0.00e+00
minmax    -0.067776058533         0.067776058533          0.00e+00   0.00e+00
```

One scalar per side, identical across all three values of `method`, constant across test points. For contrast, `CrossConformalRegressor(method="plus")` on the same data gives per-sample offsets `[-1.7288, -1.7397, -1.6993, -1.7164, ...]`.

So `method` selects how the fold quantile functions are aggregated before a single pooled out-of-fold quantile is added to them, and does not enter the conformal correction at all. Two questions rather than a conclusion, since the intent is yours to set:

1. Is that the intended estimator? If so, this docstring reads as if the three options were different conformalization strategies, which I think will mislead people.
2. Which result is the guarantee from? The `References` block below cites Romano et al. 2019, which covers split CQR. Worth noting that `"plus"` is documented differently for the sibling class — `mapie/regression/regression.py:371-372` describes it as using "the conformity scores from each fold **and the test set predictions**" — so the same keyword means two constructions across the two estimators.

mapie/regression/quantile_regression.py
```diff
@@ -1,34 +1,1436 @@
...
+                cross-validation fold
+            - "median": Aggregates (using median) the predictions of the regressors
+                trained on each cross-validation fold
+            - "pinball_weighted_mean": Aggregates the predictions of the regressors
+                trained on each cross-validation fold using a weighted mean, where the
+                weights are the pinball losses of each fold.
+
+            The three aggregations above are invalid with `method="base"`.
+
+        minimize_interval_width : bool, default=False
```
shivamlalakiya, line 1233: **Suggestion.** `minimize_interval_width` is unreachable in the default configuration. The default `conformity_score=AbsoluteQuantileRegressionScore` is `sym=True`, and `get_bounds` rejects that pairing at `mapie/conformity_scores/regression.py:412-416`:

```
ValueError: Interval width minimization cannot be used with a symmetrical conformity score function.
```

Either say here that it requires an asymmetric score, or raise earlier with a message that names the fix.

mapie/conformity_scores/bounds/quantile.py
```diff
@@ -0,0 +1,102 @@
...
+
+        return np.add(y_pred, conformity_scores)
+
+
+class AbsoluteQuantileRegressionScore(QuantileRegressionScore):
+    """
+    Absolute conformity score for quantile regression.
+    """
+
+    def __init__(
```
shivamlalakiya, line 79: **Suggestion.** `sym` is settable here, and `_check_score` (`mapie/regression/quantile_regression.py:179-190`) accepts any `QuantileRegressionScore` subclass, so `AbsoluteQuantileRegressionScore(sym=False)` passes. The score is then a single max, but `get_bounds` takes its `beta` quantile for the lower side and *adds* it, rather than subtracting a `1-alpha` quantile. Measured on n=200 calibration, 20000 fresh points, nominal 0.90:

```
AbsoluteQuantileRegressionScore()          coverage=0.9060  width=3.4872   (widened below +0.0678, above +0.0678)
AbsoluteQuantileRegressionScore(sym=False) coverage=0.9721  width=5.4632   (widened below +1.7630, above +0.3485)
```

Conservative rather than under-covering, so not blocking — but 57% wider for no reason, and the one-sided asymmetry is an artefact of the branch, not of the data. Cheapest fix is to drop `sym` from this signature, or reject `sym=False` here.

HISTORY.md
```diff
@@ -24,6 +24,13 @@
 - Raise `NotImplementedError` from the conditional conformal estimators when the infinite-dimensional (RKHS) component is requested via `infinite_params`; the supporting code is retained for future work.
```
shivamlalakiya, line 24: **Suggestion.** This diff is +27/-0 and all 27 added lines duplicate entries already in the file — each of the 27 now appears exactly twice. Among them: ``Add `StdConformityScore` for regression models…``, ``Consolidate quantile computation: extract `_compute_regression_quantile`…``, and ``Add optional `covmetrics` integration for conditional coverage diagnostics.`` Looks like a merge went sideways.

Meanwhile `CrossConformalizedQuantileRegressor` appears nowhere in the file, though the PR checklist marks HISTORY as updated.

mapie/regression/quantile_regression.py
```diff
@@ -1,34 +1,1436 @@
...
+        NDArray of shape (n_samples_test, 1)
+            The predictions.
+        """
+        check_is_fitted(self)
+
+        n_samples = _num_samples(X)
+        n_splits = self.cv.get_n_splits(X, y, groups)
+        indices = [calib_index for _, calib_index in self.cv.split(X, y, groups)]
+        if len(self.n_calib_samples) < n_splits:
+            self.n_calib_samples = [len(calib_index) for calib_index in indices]
```
shivamlalakiya, line 710: **Suggestion.** `n_calib_samples` is dead on `_QuantileConformalizer` — initialised at `:336` and `:522`, assigned here, and never read. The pooled out-of-fold layout means the per-fold sizes never reach the quantile.

Scoping this carefully because the attribute name is shared: the reads at `:2194`, `:2197`, `:2203`, `:2315`, `:2317` belong to `_MapieQuantileRegressor` (`:1728`) and are live — leave those alone. Worth deleting only the `_QuantileConformalizer` copies, so nobody later assumes the calibration count is per-fold.

mapie/tests/test_quantile_regression.py
```diff
@@ -591,13 +672,1964 @@ def early_stopping_monitor(i, est, locals):
...
+
+    shifted = scores.copy()
+    shifted[1] += 5.0
+    shifted_low, shifted_up = _asymmetric_bounds(shifted)
+
+    assert shifted_up > bound_up
+    assert shifted_low == bound_low
+
+
+def test_asymmetric_bounds_split_the_miscoverage_between_both_sides() -> None:
```
shivamlalakiya, line 2494: This is the right kind of test and I want to say so explicitly: it asserts miscoverage against an actual distribution rather than against a re-run of the implementation's own arithmetic. Grepping this file for `np.quantile|nanquantile|np.ceil|_compute_regression_quantile` gives zero hits, so no expected value here is produced by the code under test. That is the anti-pattern that let `_compute_classification_quantile` carry an off-by-one for years.

**The gap is range, not style.** Every calibration set in the new tests sits well outside the silent-clip windows in my comment on `mapie/regression/quantile_regression.py:1271`: `_asymmetric_scores` uses 500, this test 5000, the end-to-end tests `X[:100]`. That is why the guard fix changes no test result across +2118 lines added to this file. A test inside the window would close it:

```python
def test_asymmetric_score_rejects_a_calibration_set_too_small_for_alpha_over_two():
    """alpha/2 needs ceil((1-alpha/2)(n+1)) <= n; n=15 at alpha=0.1 does not."""
    reg = CrossConformalizedQuantileRegressor(
        estimator=qt,
        confidence_level=0.9,
        conformity_score=QuantileRegressionScore(),
        cv=3,
        random_state=0,
    ).fit_conformalize(X[:15], y[:15])
    with pytest.raises(ValueError, match="Number of samples of the score is too low"):
        reg.predict_interval(X[:5])
```

shivamlalakiya (commented):
Outside contributor, not a maintainer. I've been through `_compute_regression_quantile` and `_compute_classification_quantile` recently (#973, #974), so I read this PR through that lens. Every finding below was reproduced by running code on the branch; nothing here is speculative.

Provenance, so the numbers are checkable: I worked from `49205b86`. Every `mapie/` file these findings touch is byte-identical at the current head `ef8fc16c` — that merge only moved docs, `exchangeability_testing/`, `risk_control/risks.py`, and one `utils.py` docstring — and I re-verified every line number quoted below against `ef8fc16c`.

Two things worth naming specifically, because this is a large piece of work with no substantive review yet. The `_Conformalizer` / `_QuantileConformalizer` seam is the right cut — `get_bounds` no longer hard-depends on `EnsembleRegressor`, and that unblocks a refactor this module has needed for a while. And the new scores keep the `y - y_pred` orientation the rest of `conformity_scores/bounds/` uses, so `get_estimation_distribution` composes with them instead of needing a special case.

One blocking correctness item, one API/docs question, four suggestions, and a note on test range — all inline, each at its anchor. No style or formatting comments.

Cross-refs: #974 for the underlying clip, #973 for the same class of off-by-one on the classification side. Neither conflicts with this branch — `git merge-tree --write-tree upstream/cross-CQR pull/973/head` merges clean, so landing order doesn't matter.

Happy to send the guard fix plus its test as a PR against this branch if that's useful — it's four lines and a test, already verified green, and it would save you a round trip.

PR #963, opened 2026-07-21
feat: Automatically Adaptive Conformal Risk Control

allglc:
# Implementation details:

## `objective_sign`

`RiskLoss` first normalizes every controlled quantity into a loss: when
`higher_is_better=True`, the returned performance metric is converted to
`1 - value`. Its `monotonicity` therefore describes this normalized loss as a
function of the prediction parameter.

The integrated AA-CRC objective has a different optimization orientation
depending on that monotonicity. For an increasing loss, the objective is
minimized as written; for a decreasing loss, the same expression must be
maximized. `RiskLoss.objective_sign` converts both cases into the minimization
convention used by PyTorch:

```python
objective_sign = 1.0 if monotonicity == "increasing" else -1.0
```

The sign is applied to the complete gradient surrogate, including the
conformalization terms and the worst-case term. This lets the optimizer use the
same code path for recall loss (increasing after converting recall to
`1 - recall`) and miscoverage loss (decreasing as interval width grows), while
custom risks declare their direction explicitly.

## `batch_scale`

The theoretical objective contains a sum over all `n` conformalization samples,
divided by `n + 1`. Training uses mini-batches of size `b`, so directly summing a
batch would shrink the conformalization contribution relative to the single
worst-case test-point term and would make the optimization depend on
`batch_size`.

Each step therefore rescales the sampled conformalization sum by

```python
batch_scale = n / b
```

and computes

```text
((n / b) * sum(batch gradients) + worst-case gradient) / (n + 1).
```

For a uniformly sampled mini-batch, `(n / b) * sum(batch gradients)` is an
unbiased estimator of the full conformalization sum. The worst-case term is not
scaled because it is one separate contribution, not a sampled part of that
sum. Using the actual current batch size also handles the shorter final batch
correctly.

## `LinearHead` not bounded by default

The default parameter model is `_LinearHead`, a single affine layer with no
sigmoid or clipping in its forward pass. Correspondingly,
`predict_param_range=None` by default, so the learned prediction parameter is
unbounded.

This choice follows the AA-CRC theory: the guarantee assumes that the learned
functions belong to a vector space. Affine/linear heads satisfy that structural
assumption, whereas bounding the output through a sigmoid or clipping does not
preserve closure under addition and scalar multiplication.

The default head is initialized to output zero everywhere (zero weights and
zero bias). Users can still provide `predict_param_range=(lower, upper)` as a
practical safeguard; in that case the final prediction parameters are clipped
to the requested interval, and the default bias is initialized at its midpoint.
This bounded variant is useful in applications but does not strictly retain the
paper's vector-space guarantee. A custom `base_model` can also be supplied when
a different parameterization is desired.

### Cause

`_compute_integrals` approximates

```text
I(u) = integral from integration_start to u of (loss(t) - alpha) dt
```

with a 100-point trapezoidal grid. Previously, the grid endpoint was the
differentiable `predict_param`, so every integration grid point moved with it.
Autograd therefore differentiated the finite-grid approximation rather than
using the endpoint identity

```text
dI/du = loss(u) - alpha.
```

This is numerically unreliable for steep losses. For example, the built-in
miscoverage loss uses a sigmoid with slope 1000, whose transition is much
narrower than the grid spacing. With residual 1, `u=2`, and `alpha=0.1`, the
correct gradient is `-0.1`, but differentiating the moving grid returned about
`+0.39917`. The integral value itself remained accurate; only its gradient was
wrong and could point optimization in the opposite direction.

### Decision

Commit `916069b9c` removed loss-value-based model selection. Training now only
calls `loss.backward()` before the optimizer step, so the numerical value of
the integrated objective is no longer consumed; only its gradient matters.

The numerical integration grid was therefore removed completely:

1. Evaluate each risk once, at its prediction-parameter endpoint, in one
   batched call.
2. Use the endpoint risk values to obtain the exact integrated-objective
   derivatives:
   `integrals_endpoint_derivatives = endpoint_losses - alpha`.
3. Multiply those derivatives by zero-valued gradient proxies:

```python
integrals_gradient_proxy = integrals_endpoint_derivatives * (
    predict_params - predict_params.detach()
)
```

The proxy is zero in the forward pass, but its derivative with respect to
`predict_params` is `integrals_endpoint_derivatives`. The worst-case term uses
the same pattern. `_AACRCLoss` consequently returns a zero-valued gradient
surrogate, not the numerical AA-CRC objective; it must not be used for model
selection or loss-value logging.

This preserves the exact optimizer gradient while eliminating the per-sample
loop, the 100-point grids, repeated targets and predictions, `torch.trapz`, and
the backward graph through the risk. The algorithm needs only endpoint risk
values, not derivatives of the risk itself.

### Validation

Regression tests reproduce the steep-miscoverage case, verify the exact
endpoint-derived optimizer gradients, confirm that endpoint risk is evaluated
once for the complete mini-batch, and cover mini-batch scaling, objective
orientation, and worst-case gradients. The complete risk-control suite passes
with 665 tests.


examples/risk_control/3-scientific-articles/plot_blot2025_aa_crc.py
GBrelurut, line 1: Je ne trouve pas l'article dans la documentation en ligne

GBrelurut (approved):
Approuvée pour ne pas bloquer le merge une fois le problème d'affichage du notebook réglé.

examples/risk_control/3-scientific-articles/plot_blot2025_aa_crc.py
allglc: C'est bien dans risk control examples /  scientific articles. Tu as bien besoin de vacances !

PR #973, opened 2026-08-02
fix: classification quantile is one order statistic too high

shivamlalakiya:
# Description

Hello, and apologies for not using the template on the first pass — filling it in properly here.

`_compute_classification_quantile` returns one order statistic more than the split-conformal rule requires, for a large family of `(n, confidence_level)` pairs.

Split conformal needs the `k`-th smallest calibration score, with `k = ceil((1 - alpha) * (n + 1))`. The current implementation is:

```python
np.quantile(conformity_scores, (n + 1) * (1 - alpha) / n, method="higher")
```

numpy's virtual index for `method="higher"` is `ceil(q * (n - 1))`. Writing `c = (1 - alpha) * (n + 1)` as evaluated in floating point, the current form returns the `ceil(c - c/n) + 1`-th smallest score while the target is the `ceil(c)`-th, so the two diverge exactly when

```
c < n   and   ( frac(c) == 0   or   frac(c) > c / n )
```

Over `n = 5..200` against `alpha` in `{0.01, 0.05, 0.1, 0.2, 0.3, 0.5}` — 1064 feasible cells, after dropping the infeasible `k > n` ones — 225 cells diverge, and that predicate agrees with the measured behaviour on all 1064. It also holds on a wider sweep of roughly 314,000 cells over ~300 values of `alpha`, with zero counterexamples. Where the two differ, the current form is always exactly one order statistic above the target, never below.

> **Correction.** An earlier version of this description said the discrepancy occurs "whenever `(1 - alpha) * (n + 1)` is an integer". That is the majority case but not the whole one, and it is wrong in both directions: it misses 39 cells with non-integer `c`, and it wrongly flags 3 cells at `c = n` — `(n, alpha) = (9, 0.1)`, `(19, 0.05)`, `(99, 0.01)` — where the virtual index saturates on `max(scores)` and both forms agree. Thanks to @LEDazzio01 for [catching this](https://github.com/scikit-learn-contrib/MAPIE/pull/973#issuecomment-5198180665).

On the score vector `1..99`, plus the smallest non-integer cell:

| n | alpha | `c` | `k = ceil(c)` | `_compute_regression_quantile` | `_compute_classification_quantile` |
|---|---|---|---|---|---|
| 99 | 0.10 | 90.0 | 90 | 90.0 | **91.0** |
| 99 | 0.05 | 95.0 | 95 | 95.0 | **96.0** |
| 99 | 0.01 | 99.0 | 99 | 99.0 | 99.0 (`c = n`) |
| 6 | 0.30 | 4.9 | 5 | 5.0 | **6.0** |

The direction of the error is conservative: prediction sets are wider than necessary by one calibration point, and coverage is never below `1 - alpha`. This is not a validity bug. It does mean that the two helpers extracted in #944 disagree with each other, while #479 asked for classification and regression to share a single quantile rule — `_compute_regression_quantile` already applies the correct `ceil()`-based correction.

The fix is the expression `_compute_regression_quantile` already uses:

```python
np.quantile(conformity_scores, np.ceil((n + 1) * (1 - _alpha)) / n, method="lower")
```

Behaviour in the infeasible region (`k > n`) is unchanged: the quantile level exceeds 1 under both the old and the new expression, so `np.quantile` still raises.

## Why the current behaviour is in the code

It is not an oversight, and @allglc should get to make this call knowingly rather than have it land as a plain bug fix.

#944 originally unified the two helpers. @allglc [asked for the classification path to stay close to the original](https://github.com/scikit-learn-contrib/MAPIE/pull/944#issuecomment-4717095367):

> `_compute_regression_quantile` does not change much the original implementation but `_compute_classification_quantile` changes a lot the original implementation. I would prefer keeping close to the original. Especially that it seems to change behavior.

so it was reverted to `method="higher"` with no `ceil()`, and the behaviour this PR changes is behaviour that was deliberately preserved. The difference from #944 is that the change is no longer incidental to a refactor: it is isolated, argued from the order statistic, and covered by tests that fail without it. @LEDazzio01, who wrote both helpers, [set out that history](https://github.com/scikit-learn-contrib/MAPIE/pull/973#issuecomment-5198180665) and independently verified the fix.

Relates to #479, follows up on #944. No public API change.

## Type of change

# How Has This Been Tested?

- [x] `test_returns_conformal_order_statistic` — asserts the returned value equals `scores[k - 1]`. It covers both families: the integer-`c` cells, and the two smallest non-integer ones, `(n, alpha) = (6, 0.3)` and `(13, 0.3)`, where `c = 4.9` and `c = 9.8`. The original grid could not reach a non-integer cell for any `n`: with `alpha` in `{0.05, 0.1, 0.2, 0.5}` none exists, so the `alpha` list had to change. `n = 6` is infeasible at 0.05 and 0.1 and `n = 13` at 0.05, so the cells are a flat list of pairs rather than a nested grid. This test replaces `test_matches_old_formula`, which asserted the implementation against a copy of its own formula and so could not detect any of this.
- [x] `test_matches_regression_quantile` — asserts that the two helpers now agree (#479), over the same two families. Both tests fail on `master` and pass with this change.
- [x] Fixtures: 14 entries over 28 literal lines in `test_classification.py` — 8 prediction-set arrays and 6 `COVERAGES` constants. All belong to `test_toy_dataset_predictions` and `test_toy_binary_dataset_predictions`, which run on `X_toy` (`n = 9`) at `alpha = 0.5`, so `c = 5` and `c < n`. I recomputed the expected LAC / APS / top-k sets from first principles rather than from the new code's output: the independent computation reproduces the old fixtures at `k = 6` and the new ones at `k = 5`.
- [x] That fixture set is complete regardless of how it was found. Every assertion consuming those fixtures compares against a pinned literal at default tolerance, so a fixture sitting on any other discrepancy cell — integer `c` or not — would have failed too and would appear in this diff. The suite is green with only these 14 changed. The one assertion in the file carrying a real tolerance, `test_large_dataset_predictions` (`rtol=1e-2`), sits at `n = 500`, `alpha = 0.2`, where `c = 400.8` and `frac(c) = 0.8000 < c/n = 0.8016` — not a discrepancy cell, so it is not absorbing a change either.
- [x] Merged current `master` (60031ac7, the v1.5.0 pre-release) into the branch, resolving the `HISTORY.md` and `AUTHORS.md` conflicts against the restructured files: the `HISTORY.md` entry for this fix now sits under the new empty `1.x.x` section, and none of the entries the pre-release moved or reworded were re-added. Merged rather than rebased so the commit hashes referenced above stay valid — happy to rebase and force-push instead if you prefer a linear history.
- [x] `make lint`, `make format` and `make coverage` pass locally — 100% branch coverage.

# Checklist

## Guidelines

## Quality Checks
- [x] Typing passes successfully: `make type-check` (green in the Quality checks job; on my machine mypy errors inside the numpy stubs for Python 3.13, unrelated to this change)
- [ ] When updating documentation: doc builds successfully and without warnings: `make doc` — n/a, no documentation changes
- [ ] When updating documentation: code examples in doc run successfully: `make doctest` — n/a, no documentation changes

## Contribution Documentation

## LLM Usage

If you used an LLM, please provide the following details:

- **LLM used**: Claude (Claude Code)
- **Purpose**: drafting this pull request description, plus mechanical repository chores (rebasing onto `master`, adding the AUTHORS.md line). The fix, the two tests and the fixture recomputation are my own work. Every number in this description was produced by executing the code, not quoted from the model.


shivamlalakiya:
Pushed c989b42 addressing the two code-scanning comments: both new assertions are exact-equality checks on order statistics, so `np.testing.assert_array_equal` preserves the semantics and clears the Bandit `B101` alerts. Verified they still fail when the change to `_compute_classification_quantile` is reverted, so they remain real regression guards.

On the failing `sonarqube` check — I don't think there is anything I can do about it from a fork. The job log shows:

```
SONAR_TOKEN:
SONAR_HOST_URL:
##[warning]Running this GitHub Action without SONAR_TOKEN is not recommended
ERROR Failed to query server version: URI with undefined scheme
```

GitHub doesn't expose repository secrets to workflows triggered by pull requests from forks, so the scanner has no host to talk to. It passes on same-repo branches (#971, #972) and did not pass on the last fork PR either (#944). Happy to rebase or adjust if you'd rather handle it differently.

Everything else is green: tests on ubuntu / macos / windows / minimum-supported, quality checks, Bandit, CodeQL, and docs. Locally `make lint`, `make format` and `make coverage` pass — 2744 passed, 100% coverage.

allglc:
Could you use the pull request template?
Please know that contributions made by agents without any added value by a human (at least a careful review before opening a PR, a human-written description with a "hello" would be nice too...) do not bring any value: it takes maintainers a long time to review, and we can use agents ourselves.

shivamlalakiya:
Hi @allglc — fair point, sorry for skipping the template. Rewrote the description above to follow it, including LLM Usage: Claude drafted the write-up and did the mechanical chores (rebase, AUTHORS.md line); the fix, the two tests, and the fixture recomputation are mine. Also rebased onto current master (6478f16).

**Why I was in this code.** I'm comparing how conformal libraries resolve the quantile *level* at small calibration sizes, across MAPIE and crepes, for a write-up. To do that, I wrote an oracle that computes `k = ⌈(1−α)(n+1)⌉` in exact rationals — independent of MAPIE's code — and swept `n = 2..500` against several α, on both quantile helpers. `_compute_regression_quantile` matched the oracle; `_compute_classification_quantile` came back one order statistic high on a subset of cells. Cause: numpy's virtual index for `method="higher"` is `⌈q(n−1)⌉`, which lands on `k+1`. Conservative, not a validity bug — hence the wording in the description. It isn't limited to integer `(1−α)(n+1)`: the general condition is `frac((1−α)(n+1)) > (1−α)(n+1)/n`, which collapses to the integer case only for unit-fraction α. At α = 1/3, 33% of feasible `n` get the extra order statistic.

**The part worth a second pair of eyes** is the fixture change, not the one-line fix. MAPIE's toy cases sit at `n = 9, α = 0.5`, where `(1−α)(n+1) = 5` exactly, so eight of them encoded the old behaviour. The same independent LAC/top-k computation reproduces the old arrays at `k = 6` and the new ones at `k = 5`.

Two CI notes, neither from this diff:
- **sonarqube** — fork PRs get no repository secrets, so `SONAR_TOKEN`/`SONAR_HOST_URL` are empty and the scanner dies on `Failed to query server version: URI with undefined scheme`. Same on #944; green on same-repo branches like #971.
- **readthedocs** — red on master too. Build 33906218 (master, 6478f16) fails with `Index file is not present in HTML output directory`; 33904668 (8a7a71c) was green. `6478f16` moved docs under `doc/text/`, so with `docs_dir: doc` and `nav: Home: text/index.md` there's no `site/index.html` for RTD's healthcheck. `mkdocs build` itself exits 0. Left alone as unrelated — happy to open a separate PR or issue if useful.

LEDazzio01:
Hi @shivamlalakiya — I wrote `_compute_classification_quantile` and `_compute_regression_quantile` in #944, so I wanted to add some context, and to flag one thing about the PR description.

To be clear about where I stand: I'm an outside contributor, not a maintainer, so this is a comment rather than a review — @allglc and @GBrelurut own the call here.

## The fix is correct

I verified it independently rather than reading the diff. Sweeping `n = 5..200` against `alpha ∈ {0.01, 0.05, 0.1, 0.2, 0.3, 0.5}` (1064 cells, after dropping the infeasible `k > n` ones) and comparing each form against the split-conformal target — the `ceil((1-α)(n+1))`-th smallest calibration score:

| form | cells disagreeing with the target |
|---|---|
| current — `((n+1)(1-α))/n`, `method="higher"` | **225** |
| proposed — `ceil((n+1)(1-α))/n`, `method="lower"` | **0** |

The proposed form also matches `_compute_regression_quantile` on all 1064 cells, which is the #479 goal. And I agree the direction is conservative: sets are wider by one calibration point, coverage is never below `1 - α`, so this isn't a validity bug.

## The stated condition is narrower than the actual bug

The description says the discrepancy occurs "whenever `(1 - alpha) * (n + 1)` is an integer." That's the common case but not the whole one. Of the 225 mismatching cells, **186 are integer cells and 39 are not**.

The smallest counterexample is `n = 6, α = 0.3`, where `c = (1-α)(n+1) = 4.9` is not an integer:

```
k = ceil(4.9) = 5     ->  target is the 5th smallest = 5.0
current: q = 4.9/6 = 0.81667, virtual index = q(n-1) = 4.0833,
         method="higher" -> ceil -> index 5 -> returns 6.0
```

Off by one, same as the integer cells. Writing `c = (1-α)(n+1)`, `method="higher"` returns the `ceil(c - c/n) + 1`-th smallest while the target is the `ceil(c)`-th, so the forms diverge exactly when

```
c < n   and   ( frac(c) == 0   or   frac(c) > c/n )
```

The `c < n` clause matters: at `c = n` the virtual index saturates on the largest order statistic and both forms return `max(scores)`. I checked this predicate against the empirical result on all 1064 cells — 0 mismatches, using exact rationals so it isn't a float artifact.

This doesn't change the fix, which is right either way. It matters for the fixture work: if the 28 recomputed cases in `test_classification.py` were selected by looking for integer cells, it's worth re-checking them against the full predicate. The toy fixtures at `n = 9, α = 0.5` give `c = 5`, so those are covered — but any fixture sitting on one of the 39 non-integer cells would also have encoded the old behaviour.

## History this PR should probably carry

The current implementation isn't an oversight — it's there because a maintainer asked for it.

In #944 I originally unified the two paths. @allglc pushed back ([comment](https://github.com/scikit-learn-contrib/MAPIE/pull/944#issuecomment-4717095367)):

> `_compute_regression_quantile` does not change much the original implementation but `_compute_classification_quantile` changes a lot the original implementation. I would prefer keeping close to the original. Especially that it seems to change behavior.

So I reverted the classification helper to match the original `_compute_quantiles` exactly — `method="higher"`, no `ceil()`. The behaviour this PR is changing is behaviour that was deliberately preserved.

That's not an argument against the change. If anything it's the opposite: the concern then was an *incidental, untested* behaviour change buried inside a refactor, and this PR is the deliberate, tested version, with the order-statistic reasoning written down and fixtures recomputed from first principles. But @allglc should get to make that call knowingly rather than have it land as a bug fix.

## Correcting myself

While I'm here — I argued in #944 that the two paths couldn't be unified ([comment](https://github.com/scikit-learn-contrib/MAPIE/pull/944#issuecomment-4648885942)):

> These produce different values for every combination of `n` and `α` I tested, so they can't be interchanged without breaking one path's coverage guarantee.

The first half is right: the two *original* formulas do differ. The conclusion I drew from it was wrong. I read "the values differ" as "both are load-bearing," and never checked which one matched the order statistic. The classification path was simply conservative — it wasn't relying on that extra order statistic for coverage. Had I checked, #944 could have unified them properly and #479 would be closed.

So the premise I used to justify splitting them doesn't hold, and this PR does what #479 actually asked for.

Thanks for digging into this — the reproducer and the from-first-principles fixture recomputation made it straightforward to check.


---

*Analysis assisted by Claude (Claude Code); I verified the numbers and reasoning.*

shivamlalakiya:
Thanks @LEDazzio01 — this is the most useful review either PR has had. You're right that my stated condition was wrong, and re-checking it turned up a second error of mine plus one refinement to your predicate. All four points below are measured against the real helper on score vectors `1..n`, so the returned value *is* the 1-based rank.

## 1. My "general condition" was wrong, and worse than the description's

In [my 4 Aug comment](https://github.com/scikit-learn-contrib/MAPIE/pull/973#issuecomment-5182336542) I wrote that the general condition is `frac(c) > c/n`. That fails in the opposite direction to the description: it picks up the non-integer family and drops every integer one, because `frac(c) = 0` does not satisfy `> c/n`. On your 1064-cell sweep:

| condition | cells disagreeing with measured behaviour |
|---|---|
| `c` integer — the PR description and HISTORY.md | 42 |
| `frac(c) > c/n` — my follow-up comment | 186 |
| `c < n and (frac(c) == 0 or frac(c) > c/n)` — yours | **0** |

Your split of the 42 is right: 39 non-integer mismatches plus 3 integer cells that don't mismatch, `(9, 0.1)`, `(19, 0.05)`, `(99, 0.01)` — all at `c = n`, the clause I'd dropped.

## 2. Your predicate is exact — but it has to be evaluated in floating point, not exact rationals

You noted you used exact rationals "so it isn't a float artifact". For your six `alpha` values the two framings agree, so nothing in your comment is affected. Beyond them they come apart, and it's the float reading that matches the code.

Sweeping `n = 5..200` against 106 values of `alpha` (every `k/100`, plus `1/3`, `1/7`, `2/3`, `1/6`, `1/9`, `0.0625`, `0.123456789`) — 20,510 feasible cells, 9,691 divergences:

| predicate evaluated in | cells disagreeing with measured behaviour |
|---|---|
| floating point | **0** |
| exact rationals | 247 |

Smallest counterexample is `n = 11, alpha = 0.45`. The code evaluates `1 - alpha` first, and that subtraction is itself a rounding: `1 - 0.45` rounds *up* to exactly `0.55`, so `c` comes out as `6.6000000000000005`, while the exact-rational value of the same expression is `6.5999999999999998667...`. The fractional parts fall on opposite sides of `c/n`:

```
float:  frac(c) = 0.6000000000000005  >  c/n = 0.6000000000000001   -> predicts divergence
exact:  frac(c) = 0.5999999999999998  <  c/n = 0.5999999999999999   -> predicts none
measured: k = 7, current form returns 8   -> the float reading is the correct one
```

So the predicate is right; it just has to model the arithmetic the code actually performs. I've stated it that way in the description.

## 3. My "33% at alpha = 1/3" figure — three different numbers, and I quoted the wrong one for the context

That claim was also in the 4 Aug comment. Measured over the 196 feasible `n` at `alpha = 1/3`:

| quantity | value |
|---|---|
| current form sits above the exact-arithmetic rank | 66 / 196 = **33.67%** |
| **this PR changes the selected order statistic** | 6 / 196 = **3.06%** (`n = 5, 11, 23, 47, 95, 191`) |
| *fixed* form still sits above the exact-arithmetic rank | 60 / 196 = **30.61%** |

I offered 33% as the impact of the bug, which reads as the impact of the fix. It isn't — that's 3.06%, an order of magnitude smaller.

The third row is the part I hadn't seen. `1 - 1/3` rounds to `0.66666666666666674 > 2/3`, so `(n + 1) * (1 - 1/3)` lands exactly on an integer only when `n + 1 = 3 * 2^k` — which is exactly `n = 5, 11, 23, 47, 95, 191`. For the other 60 of the 66 `n ≡ 2 (mod 3)`, the product is *integer + ε*, `ceil()` rounds up, and **both** the old and the new formula return that same over-shot rank. So for an `alpha` with no exact binary representation the fix corrects 6 cells and leaves 60 one order statistic above what exact arithmetic asks for.

That's conservative in the same direction, it predates both this PR and #944, and it's a property of resolving a quantile level through `float` at all — so I'm leaving it out of scope here rather than widening the diff. Worth an issue if a maintainer thinks so.

## 4. The fixture warning was justified — and the test grid was worse than the fixtures

You were right to flag the selection method, but the gap turned out to be in the *new* test rather than the fixtures.

**The fixtures are complete, and completeness doesn't depend on how I found them.** The 28 changed literal lines are 14 entries — 8 prediction-set arrays and 6 `COVERAGES` constants — all belonging to `test_toy_dataset_predictions` and `test_toy_binary_dataset_predictions`, which run on `X_toy` (`n = 9`) at `alpha = 0.5`, so `c = 5`. Every assertion consuming them compares against a pinned literal at default tolerance, so a fixture sitting on *any* other divergent cell, integer or not, would have failed too and would be in this diff. The suite is green with only these 14 changed. The one assertion in the file carrying a real tolerance, `test_large_dataset_predictions` (`rtol=1e-2`), sits at `n = 500, alpha = 0.2`, where `c = 400.8` and `frac(c) = 0.8000 < c/n = 0.8016` — not a divergent cell, so it isn't quietly absorbing anything either.

**The new test was the actual hole.** `test_returns_conformal_order_statistic` swept `n ∈ {19, 50, 99, 100, 199, 500}` against `alpha ∈ {0.05, 0.1, 0.2, 0.5}`. Of its 24 cells, 11 diverge on `master` — and all 11 have integer `c`. Its coverage of the non-integer family was zero, and no choice of `n` would have fixed that: with those four `alpha` values no non-integer divergent cell exists at any `n`. The `alpha` list had to change.

Fixed in afcb627f, which adds the two smallest non-integer cells, `(n, alpha) = (6, 0.3)` and `(13, 0.3)` (`c = 4.9` and `c = 9.8`), to both new tests. They can't be folded into the existing nested grid — `n = 6` is infeasible at `alpha = 0.05` and `0.1`, and `n = 13` at `0.05` — so the cells are now a flat list of pairs. Both tests still fail against the previous formula and pass with the fix; `make coverage` exits clean.

## 5. On #944

Agreed, and added to the description under "Why the current behaviour is in the code", quoting @allglc's comment and noting that this PR reverses that decision deliberately. @allglc — that's the part worth your attention: this is a behaviour change you previously asked to avoid, not a plain bug fix.

And thanks for the correction on your own #944 comment. From my side the useful conclusion is the same either way: the classification path was conservative rather than load-bearing, so #479 can actually be closed by making the two helpers agree.

---

*Analysis assisted by Claude (Claude Code); every number above was produced by executing the code and I verified them.*

allglc:
Thanks for the comments and the contribution!

For #944, I wanted to keep the original computation intact as it was a refactoring. However it seems indeed that there is a minor computation issue.

To avoid future maintainers to have to read everything, here is a summary of the change:

The change is necessary to make classification implement the intended split-conformal rule exactly.
method="higher" was deliberately chosen for conservative rounding, but combined with MAPIE’s finite-sample-adjusted quantile level it sometimes rounds twice, selecting the (k+1)th score instead of:
\[
k=\lceil(n+1)(1-\alpha)\rceil
\]This does not threaten coverage—it makes prediction sets unnecessarily large—but it creates avoidable conservativeness, inconsistent regression/classification behavior, and sample-size-dependent results that do not match the theoretical algorithm.
The PR keeps the required upward rounding explicitly through ceil(...); method="lower" then maps that rank correctly into NumPy’s quantile convention. Thus it changes the numerical mechanism, not the statistical rule.

